PR #13, opened 2015-07-11
Import Basic Structure

fcarstens:
- import project from https://github.com/fcarstens/js-test-examples
  - basic definition of data structures (not adjusted to latest proposal yet)
  - adapters for existing test framework APIs 
  - TAP reporter as standard API user
- restructured project in /lib and /test
- build using esperanto --> babel (rather than babel --> esperanto)

I haven't included the other adapters yet. If the structure is okay this would be my next step.
Diff view without QUnit: https://github.com/js-reporters/js-reporters/commit/65d1b17fe3be9568bd2bb51f50f24b766a81c465


.gitignore
```diff
@@ -23,3 +23,5 @@ build/Release
 # Deployed apps should consider commenting this line out:
 # see https://npmjs.org/doc/faq.html#Should-I-check-my-node_modules-folder-into-git
 node_modules
+
+.idea/
```
jzaefferer: We shouldn't add IDE or even specific entries. You can put this into a global `.gitignore` on your system, see https://help.github.com/articles/ignoring-files/#create-a-global-gitignore

dist/js-reporters.js
```diff
@@ -0,0 +1,330 @@
+(function (global, factory) {
```
jzaefferer: Let's exclude "binaries" like this, we don't want to have multiple files change in every commit, even though we change just one thing. We can include this in releases.

lib/Data.js
```diff
@@ -0,0 +1,84 @@
...
+
+var foo = new Test("foo", "passed", 42);
+var bar = new Test("bar goes wrong", "failed", 42);
+var baz = new Test("baz", "passed", 42);
+
+var groupA = new Suite("group a", [], [foo, bar]);
+var groupB = new Suite("group b", [], [baz]);
+var root = new Suite("root", [groupA, groupB], []);
+
+export var demoData = root;
```
jzaefferer: We should separate the classes defined above and this demoData. The latter shouldn't be part of the library. We can move that to the tests.

lib/Data.js
```diff
@@ -0,0 +1,84 @@
+export class Test {
```
jzaefferer: Do we need a class for this? Since there are no methods, an object literal will do just as well.

lib/Data.js
```diff
@@ -0,0 +1,84 @@
...
+    }
+
+    getAllTests(arr) {
+        if (arr === undefined) {
+            arr = [];
+        }
+        this.tests.forEach(function (test) {
+            arr.push(test);
+        });
+        this.childSuites.forEach(function (child) {
```
jzaefferer: Can you replace this with map/reduce calls, avoiding the `arr` argument?

jzaefferer:
Let's add some linting to this repo, using jshint+jscs or eslint. Preferably some preset that doesn't require any custom repo-specific rules.

lib/EventEmitter.js
```diff
@@ -0,0 +1,22 @@
...
+            this.listeners[name] = [fun];
+        }
+        else {
+            this.listeners[name].push(fun);
+        }
+    }
+
+    emit(name, ...params) {
+        if (this.listeners[name]) {
+            for (var fun of this.listeners[name]) {
```
jzaefferer: Replace `var` with `let`.

lib/EventEmitter.js
```diff
@@ -0,0 +1,22 @@
...
+                fun(...params); //callback
```
jzaefferer: Let's rename `fun` to `callback`? Also applies to `on()` above.

lib/adapters/QUnitAdapter.js
```diff
@@ -0,0 +1,73 @@
...
+        this.tests = {};
+        this.suites = [];
+    }
+
+    onTestStart() {
+        this.errors = [];
+    }
+
+    onLog(details) {
+        if (details.result != true) {
```
jzaefferer:
```js
if (!details.result) {
```

lib/adapters/QUnitAdapter.js
```diff
@@ -0,0 +1,73 @@
...
+        QUnit.moduleDone(this.onModuleDone.bind(this));
+        QUnit.log(this.onLog.bind(this));
+        QUnit.testStart(this.onTestStart.bind(this));
+
+        this.tests = {};
+        this.suites = [];
+    }
+
+    onTestStart() {
+        this.errors = [];
```
jzaefferer: This should also emit a `testStart` event. `runStart` and `suiteStart` are also missing.

lib/adapters/QUnitAdapter.js
```diff
@@ -0,0 +1,73 @@
...
+    onDone(details) {
+        var globalSuite = new Suite("", this.suites, []);
+        this.emit("runEnd", globalSuite);
+    }
+
+    onModuleDone(details) {
+        for (var test of details.tests) {
+            // check if the module is actually finished:
+            // QUnit may trigger moduleDone multiple times if it reorders tests
+            // if not, return and wait for the next moduleDone
```
jzaefferer: Add https://github.com/jquery/qunit/issues/830 as a reference to the underlying issue.

lib/reporters/TapReporter.js
```diff
@@ -0,0 +1,23 @@
...
+        if (test.status == "failed") {
+            console.log("not ok " + this.count + " " + test.testName);
+        }
+        else {
+            console.log("ok " + this.count + " " + test.testName);
+        }
+    }
+
+    onRunEnd(suite) {
+        console.log("1.." + this.count);
```
jzaefferer: We've discussed this before, just for completeness: This must be in `runStart`.

lib/Data.js
jzaefferer: Or just remove it, since it isn't used anymore.

test/qunit/index.html
```diff
@@ -0,0 +1,40 @@
...
+    <script src="../../dist/js-reporters.js"></script>
+    <link href="../../node_modules/qunitjs/qunit/qunit.css" rel="stylesheet" type="text/css"/>
+    <script src="../../node_modules/qunitjs/qunit/qunit.js"></script>
+    <script src="tests.js"></script>
+
+    <script>
+        var adapter = new JsReporters.QUnitAdapter();
+
+
+        var tapReporter = new JsReporters.TapReporter(adapter);
```
jzaefferer: This looks odd. If the resulting object isn't used at all, we shouldn't be calling a constructor. Might as well call `JsReporters.tapReporter()` or `JsReporters.TapReporter.init()`.

jzaefferer: Also mixing both custom reporter (above+below) and tap reporter seems like a bad idea, since both will output to the console.

lib/Data.js
fcarstens: Not strictly required but this way we have the data structure in one place.

lib/adapters/QUnitAdapter.js
fcarstens: Which data do we want to emit here? For example Mocha's start event does not provide any details.

lib/adapters/QUnitAdapter.js
fcarstens: (done)

lib/reporters/TapReporter.js
fcarstens: See above. It depends on the data we have at runStart.

fcarstens:
I fixed all other issues and will take a look at linting tomorrow. :smiley:

test/qunit/index.html
fcarstens: I found it useful for testing, especially because you can toggle the debug level in the chrome dev tools. I think we want automated tests at some point.

lib/Data.js
jzaefferer: Okay. Way may want to extend it later, too.

lib/adapters/QUnitAdapter.js
jzaefferer: It can include all `testEnd` properties that don't depend on actually running the test, like `suiteName`.

jzaefferer: I guess `runStart` is the more interesting event. The number of tests is obviously pretty important there. Not just for TAP, Karma also needs (at least, wants) that information.

test/qunit/index.html
```diff
@@ -0,0 +1,38 @@
...
+<html>
+<head>
+    <meta charset="utf-8">
+    <script src="../../dist/js-reporters.js"></script>
+    <link href="../../node_modules/qunitjs/qunit/qunit.css" rel="stylesheet" type="text/css"/>
+    <script src="../../node_modules/qunitjs/qunit/qunit.js"></script>
+    <script src="tests.js"></script>
+
+    <script>
+        var adapter = new JsReporters.QUnitAdapter();
```
jzaefferer: Can we refer to this as `reporter` or `qunitReporter` (or something more appropriate)? The fact that the reporter is an adapter of the QUnit API and not a native implementation shouldn't matter for its usage. The same applies to the implementation of the `TapReporter` itself.

It should become more obvious once we add a 2nd adapter, at which point we should  implement a `ConsoleReporter` and use it for both frameworks.

test/qunit/tests.js
```diff
@@ -0,0 +1,18 @@
...
+
+QUnit.module("group b");
+QUnit.test("baz", function (assert) {
+    assert.ok(true);
+});
+
+QUnit.skip("skipped test", function (assert) {
+    assert.ok(true)
+});
+    /**/
```
jzaefferer: Looks like this can be removed. Would also be good to add trailing newlines everywhere, but only after a linter is in place that complains about the lack of those.

test/qunit/index.html
jzaefferer: After some discussion with @fcarstens elsewhere: `runner` and `testRunner` (when the test context isn't obvious) should work. Once frameworks implementation the interface themselves, the adapters can be replaced, and we can refer to the frameworks object itself as the `runner`.

fcarstens:
Resolved all issues mentioned above except for runStart. Will work on that after finishing the Jasmine Adapter.

package.json
```diff
@@ -0,0 +1,29 @@
...
+    "lint": "eslint lib/"
+  },
+  "devDependencies": {
+    "babel": "^5.6.14",
+    "eslint": "^0.24.1",
+    "esperanto": "^0.7.3",
+    "jasmine-core": "^2.3.4",
+    "qunitjs": "^1.18.0"
+  },
+  "eslintConfig": {
```
jzaefferer: This doesn't seem to include any style checks by default. Let's either add those or add jscs with a preset. eslint doesn't seem to bundle any kind of preset and seems to have no plans to add any either.

package.json
```diff
@@ -0,0 +1,29 @@
+{
+  "name": "js-reporters",
+  "version": "1.0.0",
+  "scripts": {
+    "build": "esperanto -b -i lib/index.js -t umd --name JsReporters | babel -o dist/js-reporters.js --blacklist strict -m umd",
```
jzaefferer: I ran `npm install`, then `npm run build` and got this error (plus npm output):

```sh
~/dev/js-reporters (master)$ npm run build

> js-reporters@1.0.0 build /Users/jza/dev/js-reporters
> esperanto -b -i lib/index.js -t umd --name JsReporters | babel -o dist/js-reporters.js --blacklist strict -m umd

fs.js:565
fs.write = function(fd, buffer, offset, length, position, callback) {
                                                 ^
Error: ENOENT, no such file or directory 'dist/js-reporters.js'
    at Error (native)
    at Object.fs.openSync (fs.js:500:18)
    at Object.fs.writeFileSync (fs.js:1099:15)
    at output (/Users/jza/dev/js-reporters/node_modules/babel/lib/babel/file.js:79:10)
    at Socket.<anonymous> (/Users/jza/dev/js-reporters/node_modules/babel/lib/babel/file.js:97:7)
    at Socket.emit (events.js:129:20)
    at _stream_readable.js:908:16
    at process._tickCallback (node.js:355:11)
```

fcarstens: Thanks, fixed.

package.json
fcarstens: Eslint includes style checks by default, e.g. newline at the end of file: http://eslint.org/docs/rules/#stylistic-issues

fcarstens:
`mkdir -p` doesn't work on windows. I included an node package which handles this. :smiley:

lib/Data.js
```diff
@@ -0,0 +1,68 @@
...
+            failed: 0,
+            runtime: 0
+        };
+
+        var allTests = this.getAllTests();
+        allTests.forEach(function (test) {
+            if (test.status !== "passed") {
+                summary.failed++;
+            }
+            else {
```
jzaefferer: Put else on same line as closing brace, `} else {`, more below.

lib/reporters/TapReporter.js
```diff
@@ -0,0 +1,27 @@
+export default class TapReporter {
+    constructor(runner) {
+        this.runner = runner;
+        this.count = 0;
+
+        runner.on("testEnd", this.onTestEnd.bind(this));
+        runner.on("runEnd", this.onRunEnd.bind(this));
+    }
+
+    static init(runner){
```
jzaefferer: Missing space before open brace, should be `) {`.

package.json
jzaefferer: Not sure why github considers this comment outdated. Anyway: The default checks don't check for any whitespace issues. I've commented on two of those that I'd like linters to cover.

jzaefferer:
`mkdirp` looks good. Some whitespace linting and this is good to land.

fcarstens:
Done! :smiley:

jzaefferer:
Looks good! Rebased into a single commit. Looking forward to the JasmineAdapter.

test/qunit/index.html
```diff
@@ -0,0 +1,32 @@
+<!DOCTYPE html>
+<html>
+<head>
+    <meta charset="utf-8">
+    <script src="../../dist/js-reporters.js"></script>
+    <link href="../../node_modules/qunitjs/qunit/qunit.css" rel="stylesheet" type="text/css"/>
```
jzaefferer: Notice this a little late: Remove `type` attribute, remove closing slash (unnecessary on empty element), see also http://contribute.jquery.org/style-guide/html/

Should also start with `<link>` before any `<script>`.

test/qunit/index.html
```diff
@@ -0,0 +1,32 @@
...
+            console.debug("testEnd", test);
+        });
+
+    </script>
+
+</head>
+<body>
+<div id="qunit"></div>
+</body>
+</html>
```
jzaefferer: Also missing trailing LF.

Issue #4, opened 2014-09-05
Standard for test status


JamesMGreene:
In this issue thread, I would like to focus on standardizing the collection of test statuses (status names) between testing frameworks.

Some of the known status types from various frameworks:
 - **Success** &mdash; A test that succeeded without any failures, errors, or timeouts
 - **Failure** &mdash; A test that failed, typically due to an assertion failure
 - **Error** &mdash; A test that failed because of an error other than an assertion failure
 - **Timeout** &mdash; A test that failed because it took too long or sat idle for too long
 - **Skipped** &mdash; A test that was not run at all
 - **Pending** &mdash; A test that does not yet exist, i.e. is named and called out but no function/callback body is supplied
 - **Inconclusive** &mdash; A test that has been run and is **not** considered a failure despite the fact that it may have had assertion failures and/or errors. Referred to as "TODO" in TAP terminology.
 - **Global Error** &mdash; An error occurred that cannot be associated with any test

PR #15, opened 2015-07-21
JasmineAdapter and ConsoleReporter

fcarstens:
- add JasmineAdapter
- add skipped status for QUnit
- add ConsoleReporter
- add test for jasmine

### DIFF
--- a/lib/adapters/JasmineAdapter.js
+++ b/lib/adapters/JasmineAdapter.js
@@ -26,10 +26,9 @@ export default class JasmineAdapter extends EventEmitter {
     onSpecDone(details) {
         var runtime = new Date() - this.startTime;
         var skipped;
-        if (details.status === "pending"){
+        if (details.status === "pending") {
             skipped = true;
-        }
-        else {
+        } else {
             skipped = false;
         }
         var errors = details.failedExpectations;
--- a/lib/adapters/JasmineAdapter.js
+++ b/lib/adapters/JasmineAdapter.js
@@ -25,18 +25,21 @@ export default class JasmineAdapter extends EventEmitter {
 
     onSpecDone(details) {
         var runtime = new Date() - this.startTime;
-        var skipped;
+        var errors = details.failedExpectations;
+        var skipped, status;
         if (details.status === "pending") {
             skipped = true;
+            status = "skipped"
         } else {
             skipped = false;
+            status = details.status;
         }
-        var errors = details.failedExpectations;
+
 
         var test = new Test(
             details.description,
             details.fullName,
-            details.status,
+            status,
             runtime,
             errors,
             skipped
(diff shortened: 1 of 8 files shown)


lib/adapters/JasmineAdapter.js
```diff
@@ -0,0 +1,66 @@
...
+            skipped = true;
+        } else {
+            skipped = false;
+        }
+        var errors = details.failedExpectations;
+
+        var test = new Test(
+            details.description,
+            details.fullName,
+            details.status,
```
jzaefferer: This looks like its wrong at least for `skipped`, where it now ends up as `pending` instead. Also makes me wonder if the other states are consistent with the ones from the QUnit adapter.

lib/reporters/ConsoleReporter.js
```diff
@@ -0,0 +1,37 @@
+// TODO: finish grouping once suiteStart is implemented
+var hasGrouping = "group" in console && "groupEnd" in console;
+
+export default class ConsoleReporter {
+    constructor(runner) {
+        runner.on("suiteStart", this.onSuiteStart.bind(this));
```
jzaefferer: Looks like the bind calls aren't needed here, as there is no access to `this` inside any of the methods.

test/jasmine/index.html
```diff
@@ -0,0 +1,23 @@
+<!DOCTYPE html>
+<html>
+<head>
+    <meta charset="utf-8">
+    <script src="../../dist/js-reporters.js"></script>
+    <link rel="stylesheet" href="../../node_modules/jasmine-core/lib/jasmine-core/jasmine.css">
```
jzaefferer: This looks good. Can you also fix the markup in `test/qunit/index.html` accordingly?

test/jasmine/index.html
```diff
@@ -0,0 +1,23 @@
...
+    <script src="../../node_modules/jasmine-core/lib/jasmine-core/jasmine.js"></script>
+    <script src="../../node_modules/jasmine-core/lib/jasmine-core/jasmine-html.js"></script>
+    <script src="../../node_modules/jasmine-core/lib/jasmine-core/boot.js"></script>
+    <script src="tests.js"></script>
+    <script>
+
+        var runner = new JsReporters.JasmineAdapter();
+        //JsReporters.TapReporter.init(runner);
+        JsReporters.ConsoleReporter.init(runner);
```
jzaefferer: Maybe we can build some more infrastructure around this for testing purposes, where the output of both tap and console reporter are shown in an element on the page? That way we can just look at the rendered page, instead of having to check the console.

Somewhat related, we should verify that both reporters work when running inside a node process, outputting to the terminal.

lib/adapters/JasmineAdapter.js
fcarstens: Fixed, thanks. The other test states are consistent https://github.com/js-reporters/js-reporters/issues/4#issuecomment-54575585.

test/jasmine/index.html
fcarstens: Setting up a proper test infrastructure on my TODO list for next week then.

jzaefferer: Sounds good.

PR #20, opened 2015-07-26
add status to suites

### LINKED ISSUE #19: Adapters: Implement runEnd.status and testEnd.errors
Looking at the integration into browserstack-runner, the runEnd.status is pretty important to output if the run passed or failed. To output error results, I need testEnd.errors. Without those I can't really use the adapters to replace anything.

For reference, the browserstack-runner adapters: https://github.com/browserstack/browserstack-runner/tree/master/lib/_patch

fcarstens:
This PR fixes #19. The suite status is available at `globalSuite.getTotal().status`. `test.errors` should already be there?

### DIFF
--- a/lib/Data.js
+++ b/lib/Data.js
@@ -34,19 +34,30 @@ export class Suite {
         var summary = {
             passed: 0,
             failed: 0,
+            skipped: 0,
             runtime: 0
         };
 
         var allTests = this.getAllTests();
         allTests.forEach(function (test) {
-            if (test.status !== "passed") {
-                summary.failed++;
-            } else {
+            if (test.status === "passed") {
                 summary.passed++;
+            } else if (test.status === "skipped"){
+                summary.skipped++;
+            } else {
+                summary.failed++;
             }
             summary.runtime += test.runtime;
 
         });
+
+        if (summary.failed > 0) {
+            summary.status = "failed";
+        } else if (summary.passed > 0) {
+            summary.status = "passed";
+        } else {
+            summary.status = "skipped";
+        }
 
         return summary;
     }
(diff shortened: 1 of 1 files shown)


lib/Data.js
```diff
@@ -34,20 +34,31 @@ export class Suite {
...
         };
 
         var allTests = this.getAllTests();
         allTests.forEach(function (test) {
-            if (test.status !== "passed") {
-                summary.failed++;
-            } else {
+            if (test.status === "passed") {
                 summary.passed++;
+            } else if (test.status === "skipped"){
```
jzaefferer: Missing space before curly brace. Should add eslint rule for that.

jzaefferer:
I'm still not happy with the `getTotal()` method. Can we just run this inside the constructor and add the properties to the instance? Though according to our current draft, there should be `runtime`, `suites`, `tests` and `status`.

fcarstens:
This is problematic if we want to pass Test and Suite objects on the start events as well. We would then need to recalculate them somehow later on...
Alternatively, we could use es5 getters: 
```javascript 
class Suite {
  constructor(){
    this.tests = [1,2,3]; // Simple Demo
  }
  getAllTests(){
    return this.tests;
  }
  get total(){
    return this.getAllTests().length;
  }
}
Object.defineProperties(Suite.prototype, {
  toJSON: {
    value: function(){
      let ret = {};
      for(let x in this){
        ret[x] = this[x];
      }
      return ret;
    }
  },
  total: {
    enumerable: true
  }
})

var s = new Suite();
console.log(s.total, JSON.stringify(s));
// 3 {"tests":[1,2,3],"total":3}
```
What do you think?

jzaefferer:
Okay, let's give that a try.

lib/Data.js
fcarstens: My bad, didn't run lint. There's already a rule for that. Fixed in #23

fcarstens:
I implemented this on top of the test stuff: https://github.com/fcarstens/js-reporters/tree/suite_summary
Shall I open a new PR?

jzaefferer:
Maybe rebase #23 into a set of reasonably atomic commits, than add this on top?

fcarstens:
Included in #23.

PR #21, opened 2015-07-26
add license-header to dist

fcarstens:
Adds a license-header to a generated dist file. I'm not sure how we do best about the version information. Ideas? :smiley:

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
   "name": "js-reporters",
   "version": "1.0.0",
   "scripts": {
-    "build": "mkdirp dist/ && esperanto -b -i lib/index.js -t umd --name JsReporters | babel -o dist/js-reporters.js --blacklist strict -m umd",
+    "build": "mkdirp dist/ && esperanto -b -i lib/index.js -t umd --name JsReporters | babel --blacklist strict -m umd | cat lib/license-header.js - > dist/js-reporters.js",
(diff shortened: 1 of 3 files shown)


jzaefferer:
In other projects the source header has a `@VERSION` placeholder, which during the build time is replaced with the `version` property from `package.json`. Maybe using https://www.npmjs.com/package/sed.js ? Or something like that.

package.json
```diff
@@ -2,7 +2,7 @@
   "name": "js-reporters",
   "version": "1.0.0",
   "scripts": {
-    "build": "mkdirp dist/ && esperanto -b -i lib/index.js -t umd --name JsReporters | babel -o dist/js-reporters.js --blacklist strict -m umd",
+    "build": "mkdirp dist/ && esperanto -b -i lib/index.js -t umd --name JsReporters | babel --blacklist strict -m umd | cat lib/license-header.js - > dist/js-reporters.js",
```
leobalter: this is becoming too large.

I suggest splitting it to smaller steps or setting up a build file to be called here..

jzaefferer:
It looks like we're going to have ES6 JS files in `test/` as well, so may as well create a separate build script.

jzaefferer:
Still need to finish this.

fcarstens:
Superseded by #26.

PR #22, opened 2015-07-26
Fix QUnit test

fcarstens:
There is an interesting bug with QUnit at the moment. QUnit defines a `module` function which tricks our umd wrapper to believe that CommonJS is present.  The workaround is to include the reporter before QUnit. I'm not really satisfied with that but I don't have a better solution. Any ideas?

### DIFF
--- a/test/qunit/index.html
+++ b/test/qunit/index.html
@@ -3,8 +3,8 @@
 <head>
     <meta charset="utf-8">
     <link rel="stylesheet" href="../../node_modules/qunitjs/qunit/qunit.css" />
-    <script src="../../node_modules/qunitjs/qunit/qunit.js"></script>
     <script src="../../dist/js-reporters.js"></script>
+    <script src="../../node_modules/qunitjs/qunit/qunit.js"></script>
     <script src="tests.js"></script>
     <script>
 


jzaefferer:
I don't yet know where this line is coming from:

```js
typeof exports === 'object' && typeof module !== 'undefined' ? module.exports = factory() : typeof define === 'function' && define.amd ? define(factory) : global.JsReporters = factory();
```

It really should be checking for `module.exports`, not just `module`. Can we get that fixed instead?

leobalter:
If I'm not wrong, this is generated by esperanto or babel.

csnover:
Incidentally, changing that line to check for `module.exports` will probably cause different things to break if you load them into the browser-side of Electron, since it does not expose a global `exports` but *does* expose `module` and `module.exports` for the HTML file. :\

fcarstens:
The wrapper is generated by esperanto. Where do we want to fix this? I can try to work around this in the build process...

jzaefferer:
Look for open issues on the esperanto repo, if there is none, open one. Meanwhile we can try to work around it on our end.

jzaefferer:
@fcarstens have you looked into open esparanto issues or filing one yourself?

PR #23, opened 2015-07-30
Automated test infrastructure

fcarstens:
This PR tries to set up a better test infrastructure and fixes a lot of bugs on the way. I accidentally included the qunit_umd_issue  and issue_19 branches, hope that's okay. :smiley: 
- unify QUnit and Jasmine test examples
- add TestReporter which compares a runner's output to reference output
- add reference output for Jasmine and QUnit
- fix various bugs
- add test script which runs the Jasmine tests in node (QUnit tbd)
- Automated tests with Travis: https://travis-ci.org/fcarstens/js-reporters/builds  
Can we enable this for the main repo as well?


.travis.yml
```diff
@@ -0,0 +1,5 @@
+language: node_js
+before_script: "npm run build"
+script:
+  - "npm test"
+  - "npm run lint"
```
jzaefferer: Missing EOF newline. Should also add `sudo: false` to faster builds.

lib/reporters/TestReporter.js
```diff
@@ -0,0 +1,142 @@
+import {Test, Suite} from "../Data.js";
```
jzaefferer: This doesn't look like it belongs into `lib/`.

lib/reporters/TestReporter.js
```diff
@@ -0,0 +1,142 @@
...
+                this.ok = false;
+                console.error("expected:", expectedEvent[1], "actual:", data);
+            }
+        } else {
+            this.ok = false;
+            console.error("expected:", expectedEvent[0], "actual:", event);
+        }
+    }
+
+    equals(expected, actual) {
```
jzaefferer: This is very meta... the data structures above aren't exactly obvious either. This will take me a while to go through - maybe we can review this on a call?

test/test.js
```diff
@@ -0,0 +1,21 @@
+require("babel/polyfill"); // Travis runs an old version of node.
```
jzaefferer: We can specify node versions in `.travis.yml`, 0.12 should work fine.

.travis.yml
fcarstens: `sudo: false` is enabled by default: http://docs.travis-ci.com/user/workers/container-based-infrastructure/
> For repos we recognize on or after 2015-01-01, linux builds are sent to our container-based infrastructure.

jzaefferer: Good catch, no need to add that then.

lib/reporters/TestReporter.js
fcarstens: Agreed. What about the `TestReporter` class? I think it's quite useful to have this in `lib/`: next to testing our own adapters, this can be used to test if an external test framework implements the JsReporters standard correctly. Maybe we should just move the reference data?

jzaefferer: >  this can be used to test if an external test framework implements the JsReporters standard correctly

That sounds good to me. I'm not really sure how that would work though. I guess we can leave it here for now, then try to apply this to a runner, e.g. PR against Mocha that includes the test.

lib/reporters/TestReporter.js
fcarstens: Good point. I added explanatory comments to the `TestReporter`. Do you still want to review this on a call or do you think it's sufficiently less meta now?

fcarstens:
I adressed all issues mentioned above :smiley:

test/test.js
```diff
@@ -0,0 +1,20 @@
...
+    ]
+});
+
+var runner =  new JsReporters.JasmineAdapter(jasmine);
+var testReporter = new JsReporters.TestReporter(runner, referenceData.Jasmine);
+
+jasmine.execute();
+if (!testReporter.ok){
+    process.exit(1);
+}
```
leobalter: why not conditioning the exit argument to the value of `testReporter.ok`?

lib/Data.js
```diff
@@ -55,13 +66,13 @@ export class Suite {
...
-/**/
+ export class Error {
+ constructor(name, message, stack, expected, actual){
+ this.name = name;
+ this.message = message;
+ this.stack = stack;
+ this.expected = expected;
+ this.actual = actual;
+ }
+ }
```
leobalter: is this indentation intended?

lib/Data.js
```diff
@@ -34,20 +34,31 @@ export class Suite {
...
             summary.runtime += test.runtime;
 
         });
 
+        if (summary.failed > 0) {
+            summary.status = "failed";
+        } else if (summary.passed > 0) {
+            summary.status = "passed";
+        } else {
+            summary.status = "skipped";
```
leobalter: I might have missed the proper discussion, but it is safe to say a test without failing or passing shall always be treated as "skipped"?

If this was discussed before, please ignore my comment.

jzaefferer: That's a good catch. Only explicitly skipped tests can be considered skipped. Otherwise, if there are no failures, the default should be "passed".

lib/reporters/TestReporter.js
```diff
@@ -0,0 +1,93 @@
...
+    /**
+     * Gets called on each event emitted by the runner. Checks if the actual event matches the expected event.
+     */
+    onEvent(eventName, eventData) {
+        var [expectedEventName, expectedEventData] = this.referenceData.shift();
+
+        if (expectedEventName === eventName && this.equals(expectedEventData, eventData)) {
+            //event is as expected
+            if (this.ok === undefined && this.referenceData.length === 0) {
+                this.ok = true;
```
jzaefferer: Maybe initialize `this.ok` with `true`, then only ever set it to `false`? That way you don't need to check its current state.

lib/reporters/TestReporter.js
```diff
@@ -0,0 +1,93 @@
...
+
+    /**
+     * Helper function to compare
+     *  - two Test objects
+     *  - two Suite objects
+     *  - two arrays of Test or Suite objects
+     *  The equality check is not completely strict, e.g. the runtime of a Test does not have to be equal.
+     * @returns {boolean}: true if both objects are equal, false otherwise
+     */
+    equals(expected, actual) {
```
jzaefferer: Let's switch the argument order to match QUnit's and [node assert's `equal` method](https://nodejs.org/api/assert.html#assert_assert_equal_actual_expected_message), and rename to `equal`.

lib/reporters/TestReporter.js
```diff
@@ -0,0 +1,93 @@
...
+            if (!this.equals(expected.childSuites, actual.childSuites)) {
+                return false;
+            }
+
+            if (!this.equals(expected.tests, actual.tests)) {
+                return false;
+            }
+
+        } else if (expected instanceof Test) {
+            for (let x of ["testName", "suiteName", "status", "skipped"]) {
```
jzaefferer: Since we're dropping `skipped` as a property, we shouldn't include that here anymore.

jzaefferer: Maybe use `property` as the variable name, instead of `x`?

lib/Data.js
jzaefferer: Seems fine to ignore, we can address that once we get to implementing `test.errors`.

test/test.js
jzaefferer: Like this?

```js
process.exit(testReporter.ok ? 0 : 1);
```

lib/reporters/TestReporter.js
jzaefferer: And add a method that can be called at the end, it does the `referenceData.length` check and returns the appropriate boolean:
```js
done() {
  return this.ok && this.referenceData.length === 0;
}
```
Then call that instead of accessing `ok` directly.

jzaefferer:
I've now reviewed `TestReporter` properly. Let me know when you've updated this PR.

lib/Data.js
fcarstens: So if we have a suite without any tests it should be "passed" ?

jzaefferer: Yes. Should put that into the status enum draft.

lib/adapters/JasmineAdapter.js
```diff
@@ -25,24 +24,25 @@ export default class JasmineAdapter extends EventEmitter {
...
-            skipped = true;
-            status = "skipped"
+            status = "skipped";
         } else {
-            skipped = false;
             status = details.status;
         }
 
-
+        if (suiteName.lastIndexOf(testName) === suiteName.length - testName.length) {
```
jzaefferer: This deserves a code comment. From the code alone I can't tell what this does.

lib/reporters/ConsoleReporter.js
```diff
@@ -31,7 +31,7 @@ export default class ConsoleReporter {
     }
 
     onRunEnd(globalSuite) {
-        console.log("runEnd", globalSuite, globalSuite.getTotal());
+        console.log("runEnd", globalSuite, "status:", globalSuite.status, "runtime:", globalSuite.runtime);
```
jzaefferer: Doesn't the 2nd argument include both status and runtime properties?

lib/reporters/TestReporter.js
```diff
@@ -0,0 +1,92 @@
...
+        runner.on("runEnd", this.onEvent.bind(this, "runEnd"));
+    }
+
+    /**
+     * Gets called on each event emitted by the runner. Checks if the actual event matches the expected event.
+     */
+    onEvent(eventName, eventData) {
+        var [expectedEventName, expectedEventData] = this.referenceData.shift();
+
+        if (expectedEventName !== eventName || !this.equal(eventData, expectedEventData)) {
```
jzaefferer: Tiny detail, but lets switch the order of the first comparison (eventName) to be consistent with the equal call.

lib/reporters/TestReporter.js
```diff
@@ -0,0 +1,92 @@
...
+                return false;
+            }
+
+            if (!this.equal(actual.tests, expected.tests)) {
+                return false;
+            }
+
+        } else if (expected instanceof Test) {
+            for (let property of ["testName", "suiteName", "status"]) {
+                if (expected[property] !== actual[property]) {
```
jzaefferer: As above, switch order in comparison

lib/reporters/TestReporter.js
```diff
@@ -0,0 +1,92 @@
...
+                    return false;
+                }
+            }
+            if (typeof actual.runtime !== "number") {
+                return false;
+            }
+
+            if (expected.errors.length !== actual.errors.length) {
```
jzaefferer: Same here, more below

test/qunit/index.html
```diff
@@ -3,10 +3,11 @@
 <head>
     <meta charset="utf-8">
     <link rel="stylesheet" href="../../node_modules/qunitjs/qunit/qunit.css" />
-    <script src="../../node_modules/qunitjs/qunit/qunit.js"></script>
     <script src="../../dist/js-reporters.js"></script>
+    <script src="../../node_modules/qunitjs/qunit/qunit.js"></script>
```
jzaefferer: If we're landing this change, include an HTML comment to explain the order.

test/referenceData.js
```diff
@@ -0,0 +1,62 @@
...
+var baz = new Test("baz", "group b", "passed", 1, [], false);
+var skipped = new Test("skipped test", "group b", "skipped", 1, []);
+var subtest = new Test("subtest", "group with subgroup subgroup", "passed", 1, []);
+
+var groupA = new Suite("group a", [], [foo, bar]);
+var groupB = new Suite("group b", [], [baz, skipped]);
+var subgroup = new Suite("subgroup", [], [subtest]);
+var parent = new Suite("group with subgroup", [subgroup], []);
+
+// Globale Suite for QUnit
```
jzaefferer: "Global", no e at the end

test/referenceData.js
```diff
@@ -0,0 +1,62 @@
...
+    runParent,
+    [["runEnd", globalSuiteNesting]]
+);
+
+exports.QUnit = [].concat(
+    //[["runStart", globalSuiteNoNesting]],
+    runGroupA,
+    runGroupB,
+    [["runEnd", globalSuiteNoNesting]]
+);
```
jzaefferer: Missing linefeed

test/test.js
jzaefferer: Bump

lib/reporters/ConsoleReporter.js
fcarstens: Yes, this was convenience only.

fcarstens:
All fixed.

jzaefferer:
This looks good to land - can you squash the whole thing into one commit with a [proper commit message](http://contribute.jquery.org/commits-and-pull-requests/#commit-guidelines)? Speaking of, we could add [commitplease](https://www.npmjs.com/package/commitplease).

fcarstens:
All done! :smiley:

jzaefferer:
When merging a squashed commit like that, use "Closes #xxx" to reference the PR and automatically close it.

PR #24, opened 2015-08-09
Add start events

fcarstens:
This PR does the following:
- add `runStart`, `suiteStart`, `testStart` events
- refactor `JasmineAdapter` to emit start events during runtime
- refactor `QUnitAdapter` to emit start events during runtime
- adjust tests

Open issues: 
- Jasmine catches exceptions in `afterAll` hooks  and exposes them as `failedExpectations` on a suite. (How) do we want to handle this?

### DIFF
--- a/lib/Data.js
+++ b/lib/Data.js
@@ -9,6 +9,7 @@ export class Test {
 }
 
 export class Suite {
+
     /**
      *
      * @param name
@@ -33,19 +34,17 @@ export class Suite {
         var runtime = this.getAllTests()
             .map((x) => x.runtime)
             .reduce((sum, x) => sum + x, 0);
+
         // If one of the tests has not been run yet, its runtime is undefined.
         // If we add undefined and a number we end up with NaN.
-        if (isNaN(runtime)) {
-            return undefined;
-        } else {
-            return runtime;
-        }
+        return isNaN(runtime) ? undefined : runtime;
     }
 
     get status() {
         var passed = 0, failed = 0, skipped = 0;
 
         for (let test of this.getAllTests()) {
+
             // If a suite contains a test whose status is still undefined,
             // there is no final status for the suite as well.
             if (test.status === undefined) {
--- a/lib/Data.js
+++ b/lib/Data.js
@@ -33,6 +33,8 @@ export class Suite {
         var runtime = this.getAllTests()
             .map((x) => x.runtime)
             .reduce((sum, x) => sum + x, 0);
+        // If one of the tests has not been run yet, its runtime is undefined.
+        // If we add undefined and a number we end up with NaN.
         if (isNaN(runtime)) {
             return undefined;
         } else {
--- a/lib/Data.js
+++ b/lib/Data.js
@@ -30,22 +30,32 @@ export class Suite {
     }
 
     get runtime() {
-        return this.getAllTests()
+        var runtime = this.getAllTests()
             .map((x) => x.runtime)
             .reduce((sum, x) => sum + x, 0);
+        if (isNaN(runtime)){
+            return undefined
+        } else {
+            return runtime;
+        }
     }
 
     get status() {
         var passed = 0, failed = 0, skipped = 0;
-        this.getAllTests().forEach(function (test) {
-            if (test.status === "passed") {
+
+        for (let test of this.getAllTests()) {
+            // If a suite contains a test whose status is still undefined,
+            // there is no final status for the suite as well.
+            if (test.status === undefined) {
+                return undefined;
+            } else if (test.status === "passed") {
                 passed++;
             } else if (test.status === "skipped") {
                 skipped++;
             } else {
                 failed++;
             }
-        });
+        }
 
         if (failed > 0) {
             return "failed";
--- a/lib/Data.js
+++ b/lib/Data.js
@@ -33,8 +33,8 @@ export class Suite {
         var runtime = this.getAllTests()
             .map((x) => x.runtime)
             .reduce((sum, x) => sum + x, 0);
-        if (isNaN(runtime)){
-            return undefined
+        if (isNaN(runtime)) {
+            return undefined;
         } else {
             return runtime;
         }
--- a/lib/adapters/JasmineAdapter.js
+++ b/lib/adapters/JasmineAdapter.js
@@ -13,23 +13,24 @@ export default class JasmineAdapter extends EventEmitter {
             suiteDone: this.onSuiteDone.bind(this),
             jasmineDone: this.onJasmineDone.bind(this)
         });
-        // a stack of test arrays. The top element on the stack is the currently active suite
-        this.tests = [[]];
-        // a stack of suite arrays
-        this.suites = [[]];
 
+        // Map internal jasmine ids to standardized test/suite objects.
         this.idObjectMap = {};
     }
 
+    /**
+     * Recursively converts a jasmine suite into a standardized suite.
+     */
     convertSuite(jasmineSuite) {
         var name = jasmineSuite.description;
         var childSuites = [];
         var tests = [];
-        for (var i = 0; i < jasmineSuite.children.length; i++) {
-            if (jasmineSuite.children[i].id.startsWith("suite")) {
-                childSuites.push(this.convertSuite(jasmineSuite.children[i]));
-            } else if (jasmineSuite.children[i].id.startsWith("spec")) {
-                tests.push(this.convertSpec(jasmineSuite.children[i]));
+        for (let child of jasmineSuite.children) {
+            // a child can be either a suite or a spec. We use the id to distinguish between the two.
+            if (child.id.startsWith("suite")) {
+                childSuites.push(this.convertSuite(child));
+            } else if (child.id.startsWith("spec")) {
+                tests.push(this.convertSpec(child));
             }
         }
         var suite = new Suite(name, childSuites, tests);
@@ -39,12 +40,13 @@ export default class JasmineAdapter extends EventEmitter {
         return suite;
     }
 
-
+    /**
+     * Converts a jasmine spec into a standardized test.
+     */
     convertSpec(jasmineSpec) {
         var testName = jasmineSpec.description;
         var suiteName = jasmineSpec.result.fullName;
-        var status;
-        var runtime = undefined;
+        var status, runtime;
         var errors = [];
 
         if (jasmineSpec.markedPending) {
@@ -52,6 +54,8 @@ export default class JasmineAdapter extends EventEmitter {
         } else {
             status = undefined;
         }
+        // Jasmine only provides a concatenated suiteName + testName string.
+        // Remove the testName from the string.
         if (suiteName.lastIndexOf(testName) === suiteName.length - testName.length) {
             suiteName = suiteName.substr(0, suiteName.length - testName.length - 1);
         }
@@ -63,6 +67,7 @@ export default class JasmineAdapter extends EventEmitter {
             runtime,
             errors);
 
+        this.idObjectMap[jasmineSpec.id] = test;
 
         return test;
     }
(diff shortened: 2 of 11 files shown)


lib/Data.js
```diff
@@ -30,22 +30,32 @@ export class Suite {
     }
 
     get runtime() {
-        return this.getAllTests()
+        var runtime = this.getAllTests()
             .map((x) => x.runtime)
             .reduce((sum, x) => sum + x, 0);
+        if (isNaN(runtime)) {
```
jzaefferer: Under what condition do we end up with NaN? Can we detect that condition earlier?

lib/adapters/JasmineAdapter.js
```diff
@@ -4,68 +4,115 @@ import {Test, Suite} from "../Data.js";
...
             suiteDone: this.onSuiteDone.bind(this),
             jasmineDone: this.onJasmineDone.bind(this)
         });
-        // a stack of test arrays. The top element on the stack is the currently active suite
-        this.tests = [[]];
-        // a stack of suite arrays
-        this.suites = [[]];
+
+        // Map internal jasmine ids to standardized test/suite objects.
+        this.idObjectMap = {};
```
jzaefferer: This needs a better name. "object" doesn't tell the reader anything useful.

lib/adapters/JasmineAdapter.js
```diff
@@ -4,68 +4,115 @@ import {Test, Suite} from "../Data.js";
...
+        var name = jasmineSuite.description;
+        var childSuites = [];
+        var tests = [];
+        for (let child of jasmineSuite.children) {
+            // a child can be either a suite or a spec. We use the id to distinguish between the two.
+            if (child.id.indexOf("suite") === 0) {
+                childSuites.push(this.convertSuite(child));
+            } else if (child.id.indexOf("spec") === 0) {
+                tests.push(this.convertSpec(child));
+            }
```
jzaefferer: There is no else here. Can the children property contain any properties that are neither suite nor spec?

lib/adapters/JasmineAdapter.js
```diff
@@ -4,68 +4,115 @@ import {Test, Suite} from "../Data.js";
...
+
+        this.idObjectMap[jasmineSpec.id] = test;
+
+        return test;
+    }
+
+
+    onJasmineStarted() {
+        var startSuite = this.convertSuite(this.jasmine.topSuite());
+        startSuite.name = "";
```
jzaefferer: Why an empty string? Maybe it should be undefined instead?

lib/adapters/JasmineAdapter.js
```diff
@@ -4,68 +4,115 @@ import {Test, Suite} from "../Data.js";
...
-            this.tests[this.tests.length - 1].push(new Test("afterAll", "failed", 0));
-        }
-        var suite = new Suite(details.description, this.suites.pop(), this.tests.pop());
-        this.emit("suiteEnd", suite);
-        this.suites[this.suites.length - 1].push(suite);
+
+        this.emit("suiteEnd", this.idObjectMap[details.id]);
+
+        /*
+         FIXME: Jasmine Suite may have failedExpectations caused by a failure in afterAll.
```
jzaefferer: This seems fair to ignore in the adapter, but document it as a limitation and point it out to the Jasmine team.

lib/adapters/QUnitAdapter.js
```diff
@@ -5,18 +5,52 @@ import {Test, Suite} from "../Data.js";
 export default class QUnitAdapter extends EventEmitter {
     constructor() {
         super();
-        QUnit.done(this.onDone.bind(this));
+        QUnit.config.reorder = false;
```
jzaefferer: This is problematic. Is it not enough to set this for our tests here?

lib/adapters/QUnitAdapter.js
```diff
@@ -5,18 +5,52 @@ import {Test, Suite} from "../Data.js";
...
-        QUnit.log(this.onLog.bind(this));
-        QUnit.testStart(this.onTestStart.bind(this));
+        QUnit.done(this.onDone.bind(this));
+
+        this.idObjectMap = {};
+    }
+
+    convertModule(qunitModule) {
+        var name = qunitModule.name;
+        var childSuites = [];
```
jzaefferer: This is used only once, can be inlined.

lib/adapters/QUnitAdapter.js
```diff
@@ -5,18 +5,52 @@ import {Test, Suite} from "../Data.js";
...
         QUnit.moduleDone(this.onModuleDone.bind(this));
-        QUnit.log(this.onLog.bind(this));
-        QUnit.testStart(this.onTestStart.bind(this));
+        QUnit.done(this.onDone.bind(this));
+
+        this.idObjectMap = {};
+    }
+
+    convertModule(qunitModule) {
+        var name = qunitModule.name;
```
jzaefferer: This is also used just once. Can be replaced.

test/referenceData.js
```diff
@@ -19,35 +18,53 @@ var globalSuiteNoNesting = new Suite("", [groupA, groupB], []);
 // Global Suite for Jasmine
 var globalSuiteNesting = new Suite("", [groupA, groupB, parent], []);
 
+function toStartTest(test) {
+    return new Test(
+        test.testName,
+        test.suiteName,
+        undefined,
+        undefined,
+        undefined
```
jzaefferer: Explicitly passing undefined when there are no other defined arguments following seems pointless. Should just drop these three. Applies to similar usages above.

test/test.js
```diff
@@ -10,8 +13,31 @@ jasmine.loadConfig({
     ]
 });
 
-var runner =  new JsReporters.JasmineAdapter(jasmine);
-var testReporter = new JsReporters.TestReporter(runner, referenceData.Jasmine);
+var jasmineRunner = new JsReporters.JasmineAdapter(jasmine.env);
```
jzaefferer: This used to pass `jasmine` instead of `jasmine.env`. Why the change?

test/test.js
```diff
@@ -10,8 +13,31 @@ jasmine.loadConfig({
...
+var qunitTestReporter = new JsReporters.TestReporter(qunitRunner, referenceData.QUnit);
+
+QUnit.config.autorun = false;
+
+require("./qunit/tests.js");
+
+QUnit.load();
+
+done.then(function () {
+    process.exit(qunitTestReporter.ok ? 0 : 1);
```
jzaefferer: Might as well throw an exception here as well.

test/test.js
```diff
@@ -10,8 +13,31 @@ jasmine.loadConfig({
...
+jasmine.addReporter({}); // Suppress the default reporter
+var jasmineTestReporter = new JsReporters.TestReporter(jasmineRunner, referenceData.Jasmine);
 
 jasmine.execute();
-process.exit(testReporter.ok ? 0 : 1);
+if (!jasmineTestReporter.ok) {
+    throw new Error("Tests for JasmineAdapter failed!");
+}
+
+//TODO: Use a proper test framework?
```
jzaefferer: What does that mean?

test/test.js
```diff
@@ -10,8 +13,31 @@ jasmine.loadConfig({
...
+var done = new Promise(function (resolve) {
+    QUnit.done(function () {
+        resolve();
+    });
+});
+
+var qunitRunner = new JsReporters.QUnitAdapter();
```
jzaefferer: Can we pass the QUnit reference instead of relying on the global inside the adapter?

jzaefferer:
Lots of small details to improve, but overall looking good.

lib/Data.js
fcarstens: If one of the tests  has not been run yet, its runtime is undefined. If we add undefined and a number we end up with NaN. We can check all tests before, but that's probably not easier?

lib/adapters/JasmineAdapter.js
fcarstens: This is similar to Mocha which has an empty string as the global suite title. Change anyway?

lib/adapters/QUnitAdapter.js
fcarstens: If we have reordering enabled, QUnit may emit multiple module done events for one suite. I think we don't want to have this in the standard?

test/test.js
fcarstens: In node, it's `jasmine.env`, in the browser it's `jasmine.getEnv()`...

test/test.js
fcarstens: This is async, so an exception doesn't change the exit code. I added a comment to explain this.

lib/Data.js
jzaefferer: Let's add a comment than, based on your description here.

lib/adapters/JasmineAdapter.js
jzaefferer: I guess an actual name would be even better, but probably also much harder to come up with. Unless you have an idea or find a better reference, I think undefined is a better choice than the empty string.

lib/adapters/QUnitAdapter.js
jzaefferer: Let's remove this here and keep it in the tests. And create or update a QUnit issue with details of this. It's really a QUnit bug.

test/test.js
jzaefferer: An unhandled exception doesn't cause a non zero exit code? That's surprising, sounds like a bug in nodejs.

test/test.js
fcarstens: Using a test framework with proper async support would make this test easier to understand. Choosing between QUnit, Jasmine, Mocha, Buster or Intern is very meta here. Worth the effort?

lib/adapters/QUnitAdapter.js
fcarstens: Ok. Tracked at https://github.com/jquery/qunit/issues/830.

test/test.js
jzaefferer: Now I got it. Still don't understand how an unhandled exception doesn't lead to a non zero exit code. How does the assert module work then?Afaik that can be used by itself without any extra infrastructure. 

@leobalter any idea?

test/test.js
```diff
@@ -10,8 +13,33 @@ jasmine.loadConfig({
...
+    QUnit.done(function () {
+        resolve();
+    });
+});
+
+var qunitRunner = new JsReporters.QUnitAdapter(QUnit);
+var qunitTestReporter = new JsReporters.TestReporter(qunitRunner, referenceData.QUnit);
+
+QUnit.config.autorun = false;
+QUnit.config.reorder = false;
```
leobalter: This is not necessary on node envs.

test/test.js
```diff
@@ -10,8 +13,33 @@ jasmine.loadConfig({
...
+if (!jasmineTestReporter.ok) {
+    throw new Error("Tests for JasmineAdapter failed!");
+}
+
+//TODO: Use a proper test framework?
+var done = new Promise(function (resolve) {
+    QUnit.done(function () {
+        resolve();
+    });
+});
```
leobalter: This might be solved as:

```js
QUnit.done(function () {
    process.exit(qunitTestReporter.ok ? 0 : 1);
});
```

test/test.js
leobalter: I'm not sure what are we talking about, but I have some opinions from what I can see and presume:

I can't see how throwing an error on this module is better than a direct call on the process.exit if the idea is to exit. The error above on line 22 (`"Tests for JasmineAdapter failed!"`) seems weird to me.

Another issue I see on this file is that we are loading 2 reporters, one for Jasmine and one for QUnit, I wonder if they work better on individual files.

Following the file split, I'm also thinking about a proper reaction if `jasmineTestReporter.ok` or `qunitTestReporter.ok` are `false`. They can be improved to something that doesn't interrupt the tests, following to whatever comes next.

test/test.js
leobalter: Is this related to https://github.com/jquery/qunit/issues/830 ?

fcarstens: Yes, but as you said, not required on node envs. I'll remove it.

test/test.js
fcarstens: We consistently use `process.exit()` now. If we want more advanced features here, I'd suggest we use Jasmine or Mocha here (or QUnit, but Jasmine and Mocha seem to fit better for node) to structure the tests. 
Do we want to tackle this with this PR or at a later point?

fcarstens:
I addressed all comments above. Last open discussion is https://github.com/js-reporters/js-reporters/pull/24#discussion-diff-36724127.

jzaefferer:
Trying to look at that last issue again (still on mobile), the link doesn't work (outdated diff?). Can you post it again, here or on the updated diff?

lib/Data.js
```diff
@@ -30,22 +30,34 @@ export class Suite {
     }
 
     get runtime() {
-        return this.getAllTests()
+        var runtime = this.getAllTests()
             .map((x) => x.runtime)
             .reduce((sum, x) => sum + x, 0);
+        // If one of the tests has not been run yet, its runtime is undefined.
```
jzaefferer: Add blank line above line comment. See if es-lint has an option for that.

lib/Data.js
```diff
@@ -30,22 +30,34 @@ export class Suite {
...
+        // If we add undefined and a number we end up with NaN.
+        if (isNaN(runtime)) {
```
jzaefferer: How about a tertiary?

```js
return isNaN(runtime) ? undefined : runtime;
```

lib/adapters/QUnitAdapter.js
```diff
@@ -1,22 +1,52 @@
...
+        var suite = new Suite(suiteName, [], tests);
+        this.idDataMap[qunitModule.name] = suite;
+        return suite;
+    }
 
-        this.tests = {};
-        this.suites = [];
+    onBegin() {
+        var suites = [];
+        for (let suite of this.QUnit.config.modules) {
```
jzaefferer: Let's add a comment here, something like "// Access QUnit internals to build liste of modules, working around missing event data"

test/test.js
jzaefferer: I think for now its fine as is, don't need to introduce more complexity.

jzaefferer:
Did another review of the latest changes. A few more details, then this should be good land.

fcarstens:
Done! :smiley:

jzaefferer:
:+1:

Issue #12, opened 2015-05-30
 Event Data


fcarstens:
In this issue, I would like to focus on standardizing the event data format for the runEnd event. The runEnd event should provide a summary to integration tools when all tests are finished. First let's start with a look at the existing events (and please correct me if I'm wrong somewhere!).

### QUnit
`QUnit.done(function(details){}) ` provides a summary of the tests
```
{
    failed: 1, //number of assertions, not tests
    passed: 3,
    total: 4,
    runtime: 19
}
```
There is no information on passed and failed tests. 

### Mocha 
In contrast to QUnit, mocha's `end` event doesn't give information about the test data. However, we can use the global `suite end` event to get the raw data per test/suite and aggregate this for a summary:
```
{
    suites: [/* child suites (recursive) */],
    tests: [{ duration: 0, speed: "fast", state: "passed"}],
}
```
There is no information on passed and failed assertions.

### Jasmine
In Jasmine you get information on the test status only per test/spec, the total number of tests has to be counted manually during execution. The data for tests looks as follows:
```
{
    failedExpectations: [],
    passedExpectations: [],
    status: "failed" 
    // no information on current suite
}
```
In contrast to QUnit and mocha, we get information on passed and failed assertions/ expectations.

### Buster
Buster is most similar to QUnit. There's the `suite:end` event with the following data output:
```
{
    "contexts": 2, //buster context == (jasmine/mocha) suite
    "tests": 3,
    "errors": 0,
    "failures": 1,
    "assertions": 4,
    "timeouts": 0,
    "deferred": 0,
    "ok": false
}
```
### Summary
As we can see, the available event data varies greatly between the frameworks. Looking at the use-case of an integration tool, a summary as in QUnit or Buster seems to be most useful. This brings up two questions: 

1. Do we want to include information on assertions which cannot be provided by all frameworks (e.g. Mocha)?
2. Which level of detail is appropriate? Do we prefer QUnit's minimal style or do we want to provide as much information as Mocha does + summary?

jzaefferer:
I mentioned in #1 that Intern 3 implemented a new reporter interface. It looks like that has a `runEnd` event that provides an `Executor` instance. I didn't find any information about that type, the [`Custom Exectutor` docs are still a todo](https://theintern.github.io/intern/#custom-executors). May need to dig into the implementation to find out more.

[...]

PR #25, opened 2015-08-12
Add event data description to README

fcarstens:
This PR adds a description of the js-reporters event data based on #12.

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,6 +51,26 @@ The criteria for picking these event names included:
  - It uses names that are valid JavaScript identifiers, which allows using those as keys in JSON and avoids the need to quote keys in regular JS objects or function calls.
  - It doesn't overlap with any known existing events, so introducing these could be done in parallel to the existing API in each framework, optionally deprecating and eventually removing the previous API.
 
+### Event Data
+
+Based on the discussion in [#12](https://github.com/js-reporters/js-reporters/issues/12#issuecomment-120483356), there are two basic data structures:
+ - **Test**: A test holds basic information on a single test. It has the following set of required attributes
+   - `testName`: name of the test
+   - `suiteName`: name of the suite the test belongs to
+   - `status`: result of the test. Can `passed`, `failed` or `skipped`
+   - `runtime`: execution time in milliseconds
+   - `errors`: array containing all errors. Depending on the framework, this is a single exception or a list of failed assertions.
+ - **Suite**: A suite is a collection of tests and potentially other suites.
+   - `name`: name of the suite
+   - `childSuites`: array with all direct subsuites
+   - `tests`: array containing all tests that directly belong to the suite (but not to a child suite)
+   - `runtime`: execution time of the whole suite in milliseconds (including child suites)
+   - `status`: summarized status of the suite
+      - `failed`, if at least one test failed
+      - `skipped`, if all tests in the suite are skipped (and there is at least one skipped test)
+      - `passed`, otherwise (the suite may contain skipped tests, but no failed tests) 
+ 
+For `testStart` and `testEnd`, the corresponding test object is passed to the reporter.  The same applies to `suiteStart` and `suiteEnd` where the matching suite object is passed to the reporter. For `runStart` and `runEnd` a "global" suite object is passed to the reporter, which contains all top-level suites as child suites.
 
 ## Cross-Reference Issues
 
--- a/README.md
+++ b/README.md
@@ -14,7 +14,11 @@ https://github.com/js-reporters/js-reporters/issues/
 
 ## Background
 
-We on the [QUnit](http://qunitjs.com/) team have been [discussing](https://github.com/jquery/qunit/issues/531) the possibility of working with other JS test frameworks, especially those that can be run client-side (e.g. Mocha, Jasmine, Intern, Buster, etc.), to agree upon a "Common Reporter Interface" so that we could hopefully share Reporter plugins between testing frameworks. This would also benefit high-level consumers of the frameworks such as Karma, BrowserStack, SauceLabs, Testling, etc.
+We on the [QUnit](http://qunitjs.com/) team have been [discussing](https://github.com/jquery/qunit/issues/531) the
+possibility of working with other JS test frameworks, especially those that can be run client-side (e.g. Mocha, Jasmine,
+Intern, Buster, etc.), to agree upon a "Common Reporter Interface" so that we could hopefully share Reporter plugins
+between testing frameworks. This would also benefit high-level consumers of the frameworks such as Karma, BrowserStack,
+SauceLabs, Testling, etc.
 
 This would most likely come in the form of:
  - a common Reporter API/Interface, e.g.
@@ -32,7 +36,9 @@ Would _you_ be interested in discussing this with us further?  Please join in!
 
 ### Event Names
 
-Based on the discussion in [#1](https://github.com/js-reporters/js-reporters/issues/1#issuecomment-54841874), this is the suggested _minimum_ set of event names to be triggered by a testing framework, to be consumed by reporters or other testing tools.
+Based on the discussion in [#1](https://github.com/js-reporters/js-reporters/issues/1#issuecomment-54841874), this is
+the suggested _minimum_ set of event names to be triggered by a testing framework, to be consumed by reporters or other
+testing tools.
 
  - `runStart`: Indicates the beginning of a testsuite, triggered just once.
  - `suiteStart`: Triggered at the start of each group of tests within a testsuite.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -51,6 +51,26 @@ The criteria for picking these event names included:
...
 
+### Event Data
+
+Based on the discussion in [#12](https://github.com/js-reporters/js-reporters/issues/12#issuecomment-120483356), there are two basic data structures:
+ - **Test**: A test holds basic information on a single test. It has the following set of required attributes
+   - `testName`: name of the test
+   - `suiteName`: name of the suite the test belongs to
+   - `status`: result of the test. Can `passed`, `failed` or `skipped`
+   - `runtime`: execution time in milliseconds
+   - `errors`: array containing all errors. Depending on the framework, this is a single exception or a list of failed assertions.
```
jzaefferer: Add something like this: Will be empty for status other than failed.

README.md
```diff
@@ -51,6 +51,26 @@ The criteria for picking these event names included:
...
+ - **Suite**: A suite is a collection of tests and potentially other suites.
+   - `name`: name of the suite
+   - `childSuites`: array with all direct subsuites
+   - `tests`: array containing all tests that directly belong to the suite (but not to a child suite)
+   - `runtime`: execution time of the whole suite in milliseconds (including child suites)
+   - `status`: summarized status of the suite
+      - `failed`, if at least one test failed
+      - `skipped`, if all tests in the suite are skipped (and there is at least one skipped test)
+      - `passed`, otherwise (the suite may contain skipped tests, but no failed tests)
```
jzaefferer: Let's be explicit. Replace the "otherwise" and specify the exact conditions.

README.md
```diff
@@ -51,6 +51,26 @@ The criteria for picking these event names included:
...
+   - `name`: name of the suite
+   - `childSuites`: array with all direct subsuites
+   - `tests`: array containing all tests that directly belong to the suite (but not to a child suite)
+   - `runtime`: execution time of the whole suite in milliseconds (including child suites)
+   - `status`: summarized status of the suite
+      - `failed`, if at least one test failed
+      - `skipped`, if all tests in the suite are skipped (and there is at least one skipped test)
+      - `passed`, otherwise (the suite may contain skipped tests, but no failed tests) 
+ 
+For `testStart` and `testEnd`, the corresponding test object is passed to the reporter.  The same applies to `suiteStart` and `suiteEnd` where the matching suite object is passed to the reporter. For `runStart` and `runEnd` a "global" suite object is passed to the reporter, which contains all top-level suites as child suites.
```
jzaefferer: This should mention that properties can be undefined during the *start events. Or mention that while specifying each affected property.

README.md
```diff
@@ -51,6 +51,26 @@ The criteria for picking these event names included:
  - It uses names that are valid JavaScript identifiers, which allows using those as keys in JSON and avoids the need to quote keys in regular JS objects or function calls.
  - It doesn't overlap with any known existing events, so introducing these could be done in parallel to the existing API in each framework, optionally deprecating and eventually removing the previous API.
 
+### Event Data
+
+Based on the discussion in [#12](https://github.com/js-reporters/js-reporters/issues/12#issuecomment-120483356), there are two basic data structures:
```
jzaefferer: Add a bit of motivation, something like: These can represent both flat structures like used by QUnit (plus whatever else does that) as well as nested structures like (...). 

Or add this to the description for Suite?

README.md
```diff
@@ -51,6 +51,26 @@ The criteria for picking these event names included:
...
+ - **Test**: A test holds basic information on a single test. It has the following set of required attributes
```
jzaefferer: Extend this a bit to cover BDD vocabulary. Something like: In BDD, this represents a single `it()`.

jzaefferer:
In addition to the above, we should have at least two examples, one with a flat and one with a multiple levels deep nested structure. Could be addressed in a separate PR, please file an issue in that case.

fcarstens:
Included all suggestions.

README.md
```diff
@@ -14,7 +14,11 @@ https://github.com/js-reporters/js-reporters/issues/
 
 ## Background
 
-We on the [QUnit](http://qunitjs.com/) team have been [discussing](https://github.com/jquery/qunit/issues/531) the possibility of working with other JS test frameworks, especially those that can be run client-side (e.g. Mocha, Jasmine, Intern, Buster, etc.), to agree upon a "Common Reporter Interface" so that we could hopefully share Reporter plugins between testing frameworks. This would also benefit high-level consumers of the frameworks such as Karma, BrowserStack, SauceLabs, Testling, etc.
+We on the [QUnit](http://qunitjs.com/) team have been [discussing](https://github.com/jquery/qunit/issues/531) the
+possibility of working with other JS test frameworks, especially those that can be run client-side (e.g. Mocha, Jasmine,
+Intern, Buster, etc.), to agree upon a "Common Reporter Interface" so that we could hopefully share Reporter plugins
+between testing frameworks. This would also benefit high-level consumers of the frameworks such as Karma, BrowserStack,
+SauceLabs, Testling, etc.
```
jzaefferer: Why did you change everything to hard line breaks? We usually don't use those in markdown, makes it much easier to edit prose.

README.md
```diff
@@ -41,16 +47,51 @@ Based on the discussion in [#1](https://github.com/js-reporters/js-reporters/iss
  - `suiteEnd`:  Triggered at the end of each group of tests within a testsuite.
  - `runEnd`:  Indicates the end of a testsuite, triggered just once.
 
-These only define event names, not the data associated with each event. The data still needs to be specified via further discussion in [#1](https://github.com/js-reporters/js-reporters/issues/1).
+These only define event names, not the data associated with each event. The data still needs to be specified via further
+discussion in [#1](https://github.com/js-reporters/js-reporters/issues/1).
```
jzaefferer: This isn't quite right anymore, since we're adding event data below.

README.md
```diff
@@ -41,16 +47,51 @@ Based on the discussion in [#1](https://github.com/js-reporters/js-reporters/iss
...
+   [#1](https://github.com/js-reporters/js-reporters/issues/1#issuecomment-54841874)
+ - It uses names that are valid JavaScript identifiers, which allows using those as keys in JSON and avoids the need to
+   quote keys in regular JS objects or function calls.
+ - It doesn't overlap with any known existing events, so introducing these could be done in parallel to the existing API
+   in each framework, optionally deprecating and eventually removing the previous API.
+
+### Event Data
+
+Based on the discussion in [#12](https://github.com/js-reporters/js-reporters/issues/12#issuecomment-120483356), there
+are two basic data structures: Suites and Tests. A test represents an atomic test/spec/"`it()`". A suite contains tests
```
jzaefferer: I think the `it()` reads better without the quotes.

README.md
```diff
@@ -41,16 +47,51 @@ Based on the discussion in [#1](https://github.com/js-reporters/js-reporters/iss
...
+and optionally other suites. This can represent both flat structures like used by QUnit as well as nested suites like
+used by Jasmine or Mocha. The data structures are defined as follows:
+
+ - **Test**: A test holds basic information on a single test/spec. It has the following set of required attributes
+   - `testName`: name of the test
+   - `suiteName`: name of the suite the test belongs to
+   - `status`: result of the test. Can `passed`, `failed` or `skipped`
```
jzaefferer: This is redundant with the Suite status below. Let's deduplicate that, either with a separate definition that we references in both places or we define status here and reference it from below.

jzaefferer: Actually, suite status is a bit different. Maybe just explain what "skipped" is?

README.md
```diff
@@ -41,16 +47,51 @@ Based on the discussion in [#1](https://github.com/js-reporters/js-reporters/iss
...
+   in each framework, optionally deprecating and eventually removing the previous API.
+
+### Event Data
+
+Based on the discussion in [#12](https://github.com/js-reporters/js-reporters/issues/12#issuecomment-120483356), there
+are two basic data structures: Suites and Tests. A test represents an atomic test/spec/"`it()`". A suite contains tests
+and optionally other suites. This can represent both flat structures like used by QUnit as well as nested suites like
+used by Jasmine or Mocha. The data structures are defined as follows:
+
+ - **Test**: A test holds basic information on a single test/spec. It has the following set of required attributes
```
jzaefferer: Let's add either a dot or colon at the end of that sentence.

jzaefferer:
Looks good to land!

jzaefferer:
Please reference both issues and PRs in commit messages (at least before merging). Elsewhere we use "Fixes #xxx" for issues, "Closes #xxx" for PRs, could use that here as well.

PR #28, opened 2015-08-20
Adapter: add mocha adapter

fcarstens:
This PR adds an adapter for Mocha. This is semi-blocked by #24, because the start event tests aren't on master yet.

### DIFF
--- a/lib/adapters/MochaAdapter.js
+++ b/lib/adapters/MochaAdapter.js
@@ -1,0 +1,100 @@
+import EventEmitter from "../EventEmitter.js";
+import {Test, Suite} from "../Data.js";
+
+export default class MochaAdapter extends EventEmitter {
+    constructor(mocha) {
+        super();
+
+        this.mocha = mocha;
+        mocha.reporter((runner) => {
+            this.runner = runner;
+            runner.on("start", this.onStart.bind(this));
+            runner.on("suite", this.onSuite.bind(this));
+            runner.on("test", this.onTest.bind(this));
+            runner.on("fail", this.onFail.bind(this));
+            runner.on("test end", this.onTestEnd.bind(this));
+            runner.on("suite end", this.onSuiteEnd.bind(this));
+            runner.on("end", this.onEnd.bind(this));
+        });
+        this.idDataMap = {};
+
+    }
+
+    convertSuite(mochaSuite) {
+        var name = mochaSuite.title;
+        var childSuites = [];
+        var tests = [];
+
+        for (let child of mochaSuite.suites) {
+            childSuites.push(this.convertSuite(child));
+        }
+        for (let child of mochaSuite.tests) {
+            tests.push(this.convertTest(child));
+        }
+
+        var suite = new Suite(name,
+            childSuites,
+            tests
+        );
+        this.idDataMap[mochaSuite.title] = suite;
+        return suite;
+    }
+
+    convertTest(mochaTest) {
+        var testName = mochaTest.title;
+        var suiteName;
+
+        if (mochaTest.parent.title !== "") {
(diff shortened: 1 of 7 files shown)


jzaefferer:
The only thing I'm missing is additions to `test/test.js`, but that's blocked by #24, right? Since that is good to land, let me know when you've updated this, and I'll take another look.

fcarstens:
Added tests and fixed bugs. :smiley:

fcarstens:
Seems like I didn't `git add` after linting. I'm on mobile now, will fix that tomorrow. :smirk:

jzaefferer:
The next push should also build on Travis. While we've added the `.travis.yml` file, we didn't actually enable the build on Travis. That's now fixed.

lib/adapters/MochaAdapter.js
```diff
@@ -0,0 +1,113 @@
...
+    }
+
+    onFail(test, error) {
+        this.errors.push(error);
+    }
+
+    onTestEnd(details) {
+        var test = this.idDataMap[details.title];
+        if (details.pending) {
+            this.emit("testStart", test);
```
jzaefferer: Wondering why this is needed - there is no `test` event for pending tests?

test/mocha/index.html
```diff
@@ -0,0 +1,25 @@
+<!DOCTYPE html>
+<html>
+<head>
+    <meta charset="utf-8">
+    <link rel="stylesheet" href="../../node_modules/mocha/mocha.css">
+    <script src="../../node_modules/mocha/mocha.js"></script>
+    <script>mocha.setup('bdd')</script>
```
jzaefferer: Does it make any difference for the adapter what argument is passed here?

test/mocha/tests.js
```diff
@@ -0,0 +1,34 @@
...
+    it("bar goes wrong", function () {
+        assert(5 === "5");
+    });
+});
+
+describe("group b", function () {
+    it("baz", function () {
+        assert(true);
+    });
+    xit("skipped test", function () {
```
jzaefferer: Is there also `xdescribe`? Should check for that. Not quite sure what to make of this: https://github.com/mochajs/mocha/commit/6cdca19eafa34bc8f05cfa69a8084fdd6b2cf06f

jzaefferer:
Mostly looking good. Let's fix this up and land it.

fcarstens:
Thanks for reviewing! :smiley:
I have my last exam on the 30th of September and will update things afterwards. :relaxed:

jzaefferer:
@fcarstens I'm available this week to do more review, let me know when you've updated this.

test/mocha/tests.js
cynicaldevil: Maybe its wiser to have the more conventional `.skip()` test too? `xdescribe()` and `xit()` maybe deprecated in the future.
http://tjholowaychuk.tumblr.com/post/29996505898/mocha-140

jzaefferer:
You can use the code/branch from this PR as a starting point, and add commits on top. That way @fcarstens contribution can be kept intact, while we can move forward. There's no need to copy anything.

Let me know if you need help doing the above with git. This might be a good starting point, too: http://stackoverflow.com/a/14869954/2671

test/mocha/tests.js
jzaefferer: Yeah, makes sense.

cynicaldevil:
So I have to add fcartens' repo as a remote , fetch it and then start working on the mocha_adapter branch?

jzaefferer:
Yes on both!

jzaefferer:
Replaced by #36

PR #29, opened 2015-10-12
Fix Build

fcarstens:
Esperanto is now deprecated:
```
> node build.js
[DEPRECATION NOTICE] Esperanto is no longer under active development. To convert an ES6 module to another format, consider using Babel (https://babeljs.io)
[DEPRECATION NOTICE] Esperanto is no longer under active development. To bundle ES6 modules, consider using Rollup (https://github.com/rollup/rollup). See https://github.com/rollup/rollup/wiki/Migrating-from-Esperanto for help migrating
```
They pushed this as a patch release which broke our build script. :disappointed: 
Also, there is an issue with QUnit which now requires an expicit import in the test file.

### DIFF
--- a/test/qunit/tests.js
+++ b/test/qunit/tests.js
@@ -1,4 +1,10 @@
-/*global QUnit*/
+/*eslint-disable no-use-before-define*/
+if(typeof QUnit === "undefined") {
+	var QUnit = require("qunitjs");
+}
+
+/*eslint-enable no-use-before-define*/
+
 QUnit.module("group a");
 QUnit.test("foo", function (assert) {
     assert.equal(5, "5");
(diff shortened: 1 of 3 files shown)


test/qunit/tests.js
```diff
@@ -1,4 +1,10 @@
-/*global QUnit*/
+/*eslint-disable no-use-before-define*/
+if(typeof QUnit === "undefined") {
```
jzaefferer: Missing space after `if` - and missing eslint check for that.

test/qunit/tests.js
```diff
@@ -1,4 +1,10 @@
...
+	var QUnit = require("qunitjs");
```
jzaefferer: Indent here doesn't match the indent below.

jzaefferer:
Looks good. Can you fix the whitespace issues and land it?

PR #36, opened 2016-03-12
Adds mocha-adapter

cynicaldevil:
Finishes up the PR by @fcarstens from #28

### DIFF
--- a/test/mocha/index.html
+++ b/test/mocha/index.html
@@ -9,9 +9,9 @@
     <script src="tests.js"></script>
     <script>
         window.onload = function () {
-            var runner = new JsReporters.MochaAdapter(mocha);
+            // var runner = new JsReporters.MochaAdapter(mocha);
             //JsReporters.TapReporter.init(runner);
-            JsReporters.ConsoleReporter.init(runner);
+            // JsReporters.ConsoleReporter.init(runner);
 
             mocha.run();
 
--- a/test/mocha/index.html
+++ b/test/mocha/index.html
@@ -5,11 +5,6 @@
     <link rel="stylesheet" href="../../node_modules/mocha/mocha.css">
     <script src="../../node_modules/mocha/mocha.js"></script>
     <script>mocha.setup('bdd')</script>
-    <script>
-        function assert(expr, msg) {
-            if (!expr) throw new Error(msg || 'failed');
-        }
-    </script>
     <script src="../../dist/js-reporters.js"></script>
     <script src="tests.js"></script>
     <script>
--- a/test/mocha/index.html
+++ b/test/mocha/index.html
@@ -1,0 +1,30 @@
+<!DOCTYPE html>
+<html>
+<head>
+    <meta charset="utf-8">
+    <link rel="stylesheet" href="../../node_modules/mocha/mocha.css">
+    <script src="../../node_modules/mocha/mocha.js"></script>
+    <script>mocha.setup('bdd')</script>
+    <script>
+        function assert(expr, msg) {
+            if (!expr) throw new Error(msg || 'failed');
+        }
+    </script>
+    <script src="../../dist/js-reporters.js"></script>
+    <script src="tests.js"></script>
+    <script>
+        window.onload = function () {
+            var runner = new JsReporters.MochaAdapter(mocha);
+            //JsReporters.TapReporter.init(runner);
+            JsReporters.ConsoleReporter.init(runner);
+
+            mocha.run();
+
+        }
+    </script>
+
+</head>
+<body>
+<div id="mocha"></div>
+</body>
+</html>
--- a/lib/adapters/MochaAdapter.js
+++ b/lib/adapters/MochaAdapter.js
@@ -1,0 +1,100 @@
+import EventEmitter from "../EventEmitter.js";
+import {Test, Suite} from "../Data.js";
+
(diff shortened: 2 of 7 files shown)


cynicaldevil:
I forgot to `npm run test` before pushing changes. Will fix them shortly.

jzaefferer:
There's also something weird about your commit. When looking at [the commit by itself](https://github.com/cynicaldevil/js-reporters/commit/d4c466faa866eed3d47b43e7f50eb36bad1c03e8), it has [the same diff as this PR](https://github.com/js-reporters/js-reporters/pull/36/files). Not sure how that happened, but it makes the effort of commit previous commits rather usless. Specifically, I can't tell what you changed, if anything. Can you please look into that, along with fixing tests? Thanks.

jzaefferer:
PS: Please comment here when you push more commits (there are no notifications for pushes/new commits).

cynicaldevil:
Not sure why it's happening, maybe because the commit is a merge commit, and therefore it shows the diff of the two parent commits:
https://github.com/blog/559-merge-commits-are-back-and-better-than-ever

I think I only made one change, which was deleting line #92 in `lib/MochaAdapter.js`, the one which threw the `testStart` event for pending tests. I left the `mocha.setup('bdd')` as it was, because it was necessary, and I forgot to change the `xit` into `it.skip`.

Also, the index.html page in `test/mocha` did not show any output when run on the web browser. I will try to fix that too and then push commits here.

jzaefferer:
You shouldn't need a merge commit when doing a rebase. If you only changed one line, try resetting to the original branch (`git reset --hard [name of branch]`), then doing the rebase (`git rebase master`), then adding new commit with your changes.

cynicaldevil:
I reset my master and rebased, which turned out fine. 
Then, I tried to apply my change on it(deleting line #92) and then did `npm run test`. This results in some pretty weird behaviour, where the expected and actual events are off by one( I really don't know how to explain it). I can post the error output to a gist if you want to take a look at it.

jzaefferer:
I suggest you just push that state to this branch/PR, then I can try it out locally myself.

cynicaldevil:
Done.

jzaefferer:
>  This results in some pretty weird behaviour, where the expected and actual events are off by one( I really don't know how to explain it). I tried to find the reason for it, but nothing came up. I can post the error output to a gist if you want to take a look at it.

I just tried to reproduce this, but didn't get any errors. `npm test` finishes without any input, which should be okay here.

Can you gist what output you got?

PS: Landed a few commits in master, would be good to rebase this once more

cynicaldevil:
Ah, seems like I have some competition here ;)
Like I said, this happened when I tried to delete( or comment out) the `this.emit("testStart", test);` line in `MochaAdapter.js`. I could not push that change here or else the Travis build would have failed.
Here's the gist:
https://gist.github.com/cynicaldevil/d3aedde700156ea57589#file-output-js

jzaefferer:
There's a lot of output, but it looks like its mostly a off-by-one issue, where everything after the first failure is also off. May need to adjust tests to deal with that line-removal. Or maybe the line is actually correct, just needs a comment explaining why its there.

cynicaldevil:
I feel like that the line was needed, May have to do something with Mocha's internal working.

jzaefferer:
We can improve the adapter later, if we missed something. Besides the questionable line, was there anything else we should address here? If not, I'll merge this.

cynicaldevil:
There is a small issue: The HTML page which runs the tests on the browser does not display anything.

jzaefferer:
Okay. Have you tried debugging that?

cynicaldevil:
I have. I found that by commenting out the lines related to initializing the js-reporters runner object, test results of `group a` were displayed. Couldn't find anything else, but I could try fixing it if you want.

jzaefferer:
Yes please

cynicaldevil:
The HTML page now shows all the results. But as a result of the js-reporters code being commented out, the test object is not displayed in the console. If it's ok with you, then I think the code's ready to be merged.

test/mocha/index.html
```diff
@@ -0,0 +1,25 @@
...
+<head>
+    <meta charset="utf-8">
+    <link rel="stylesheet" href="../../node_modules/mocha/mocha.css">
+    <script src="../../node_modules/mocha/mocha.js"></script>
+    <script>mocha.setup('bdd')</script>
+    <script src="../../dist/js-reporters.js"></script>
+    <script src="tests.js"></script>
+    <script>
+        window.onload = function () {
+            // var runner = new JsReporters.MochaAdapter(mocha);
```
jzaefferer: I missed this earlier, let's fix this.

cynicaldevil: I found a fix, but I am not sure why it works:
```    
window.onload = function () {
    mocha.run();
    var runner = new JsReporters.MochaAdapter(mocha);
    //JsReporters.TapReporter.init(runner);
    JsReporters.ConsoleReporter.init(runner);
    mocha.run();
}
```

jzaefferer: Calling `mocha.run()` twice doesn't make sense. Can you dig into that?

cynicaldevil: I will, but I need some time. I have to study for my end sems, they're coming up next week.

jzaefferer:
Replaced by #46

PR #39, opened 2016-04-15
Refactoring testing with help of a framework

flore77:
@jzaefferer as discussed, I refactored the testing with help of a testing framework, for now Mocha and Chai. I have tried to keep it as small as possible, only to make a proof of concept, to view if it is viable and if it is worth.


jzaefferer:
This is looking promising. Can you extend this to cover another adapter? Then we'd get a better idea how much we need for each adapter, and how much code we can share for each test. It looks to me like its mostly `_getJasmineAdapterOutput` that's specifically for Jasmine.

flore77:
Yes, there is some shareable code between all adapters. I will do this, for the Qunit Adapter. Thanks for the review.

flore77:
@jzaefferer done, let me know what you think about it. 

There is also a test that is failing (the last one for the QUnit adapter), this is happening, because the 2 test fixtures are not equal, the [QUnit test fixture](https://github.com/js-reporters/js-reporters/blob/master/test/qunit/tests.js#L23) does not have nested modules (which I have seen, that they are possible) in regard to [Jasmine test fixture](https://github.com/js-reporters/js-reporters/blob/master/test/jasmine/tests.js#L22).

 I think the fixtures should be the same and have only one [referenceData object](https://github.com/js-reporters/js-reporters/pull/39/files#diff-62feaf8c8b98f77b35fd5c862d151366R48), because the output should be equal, a standardized output.

Of course there could be some exceptions, but we can test them separated for each adapter.

jzaefferer:
Nested modules are pretty new in QUnit, makes sense to updated the QUnit adapter accordingly.

Unifying the referenceData also sounds good.

I'll review the new changes/commits...

jzaefferer:
Update is looking good. Should make it relatively easy to add more frameworks.

jzaefferer:
Can we skip the nested-suite test for QUnit for now? Could then go ahead and land this.

flore77:
Yes, sure we can skip it. What do you suggest to do for now?

jzaefferer:
Yeah, this is a good start, might as well land it. The Mocha adapter (#36) could then be updated to use these tests.

flore77:
Sure, I'm ok with this. Then the whole test refactoring can be done into another issue/branch.

jzaefferer:
Can you update this, so the (new) tests actually pass?

flore77:
@jzaefferer I removed the failing test, because I couldn't do a quick fix, without changing the exiting tests, but I think the 2 remaing tests are enough to understand the idea.

jzaefferer:
Yep, that works.

PR #41, opened 2016-04-18
Using node event emitter

flore77:
@jzaefferer I switched to `Node's EventEmitter`, and everything seems ok (tests are passing).

I imported the EventEmitter the standard way, because for ES6 import style, I think we will need [rollup-plugin-node-resolve](https://github.com/rollup/rollup-plugin-node-resolve) and [rollup-plugin-commonjs](https://github.com/rollup/rollup-plugin-commonjs), but I want to know if we go this way.


jzaefferer:
That looks promising. The node implementation does quite a bit more than what we've implemented, but since that's somewhat standardized (a core module after all), it seems like a good API.

PS: Changes like this can go into a single commit.

flore77:
Yeah, I really don't know why I broke the commit in two, I cam squash them.

@jzaefferer the generated bundle will be used also in the browser? 
If yes, we still need an implementation, something like [this](https://github.com/Gozala/events) for in-browser use, I think this is also used by [webpack](https://github.com/webpack/node-libs-browser). 
`rollup` is only requiring the `events` module without importing some code, this may be fine for a node env.

jzaefferer:
Yes, used in the browser. Speaking of which, we could run our tests in Phantomjs 2.x, to test something browser-alike.

flore77:
Yes, we could have two test commands one for a `node env` and one for a `browser env`. What do you think to use for a browser env? Karma? But anyway, I think this should be a separate issue.

This issue should be if we are changing the EventEmitter implementation with a node standard compliant version.

jzaefferer:
Sure, but it helps to have tests that fail before you fix the underlying issue. Eventually we should have `npm test` run tests in at least node and Phantom.

For now let's update this to use one of the modules you suggested and manually test in a browser.

flore77:
You are right, proper TDD :smiley:

flore77:
I have really struggled to make this work. The `rollup package` was not up to date and it was not considering plugins etc. It has take me while to notice it and to get this on the right track.

As you can see the tests are passing in a node env, but in the browser I receive this error:
image

which I have seen that it is also appearing with the [old build](https://github.com/js-reporters/js-reporters/blob/43394e970cee759d0affdb1351848a4cdb815fc5/build.js#L9).

Maybe we should firstly fix the testing in another branch and then finish this.

jzaefferer:
The last commit is looking good!

> Maybe we should firstly fix the testing in another branch and then finish this.

Yeah, that makes sense.

flore77:
Alright, I will do a new branch for fixing the whole testing. And then return to this. 

Maybe the new branch would also fix #36, I have seen that there are also some problems.

jzaefferer:
Yeah, sounds like that could be related. So first fix the testing setup, then do event emitter and mocha adapter.

jzaefferer:
@flore77 let's try this again?

flore77:
@jzaefferer done.

jzaefferer:
Thanks! Squashed everything into a single commit with an updated message.

Issue #42, opened 2016-04-23
Bundling setup


flore77:
## Need

We need a proper bundle so that we can import it anywhere.

## Deliverables
A proper bundle that will also work in the browser.

## Current Problem
Now we are using `babel` to transform the [code generated](https://github.com/js-reporters/js-reporters/blob/43394e970cee759d0affdb1351848a4cdb815fc5/build.js#L17) by `rollup`, and babel is breaking the code, by passing for the `global` parameter `undefined` instead of the `this` object. The bundle looks like:

```js
(function (global, factory) {
   // some code for module importing
   global.JsReporters = factory();
})(undefined, function () {
   // some code
});
```

And of course this will not work in the browser, failing with the following error:
![image](https://cloud.githubusercontent.com/assets/11819686/14763227/137cea7e-0998-11e6-9d87-ce690f1f6f7f.png)

## Solution
To use the [rollup babel plugin](https://www.npmjs.com/package/rollup-plugin-babel), as also the [es2015 rollup babel preset](https://www.npmjs.com/package/babel-preset-es2015-rollup).

 To use this plugins we also need to update the [rollup version](https://github.com/js-reporters/js-reporters/blob/43394e970cee759d0affdb1351848a4cdb815fc5/package.json#L22).

flore77:
@jzaefferer please review this specs. 

Also please tell me:
1. do we mantain the current setup (i.e using rollup programtically), or we can switch to a `rollup.config.js` file, and then only run `rollup` from `command line`.
2. do we keep the thing with the [license-header](https://github.com/js-reporters/js-reporters/blob/43394e970cee759d0affdb1351848a4cdb815fc5/lib/license-header.js)

`rollup.config.js`:
```js
module.exports = {
  entry: 'lib/index.js',
  format: 'umd',
  moduleName: 'JsReporters',
  plugins: [
    babel()
  ],
  dest: 'dist/js-reporters.js'
};
```

jzaefferer:
I'm fine with switching to the config.js approach. We should keep the license header.

PR #43, opened 2016-04-24
Bundling Setup

flore77:
@jzaefferer please review changes, as discussed in #42.

I also changed the data format into a more friendlier format, which does not include time. Do we need also the time? Please tell me which format do you prefer :smiley:


flore77:
The tests are now also running in the browser.

jzaefferer:
Nice work. I've squashed a few related commits (with update commit message). Also updated the `npm test` command to run linting and integration tests.

PR #45, opened 2016-04-25
Standard style

### LINKED ISSUE #44: Use JavaScript Standard Style
## Need

We need a clean and organize code base.

## Deliverables

A clean code, which is easy to read.

## Solution

Use [standard](https://www.npmjs.com/package/standard), which does not need any configuration, except for ignores.

flore77:
Fixes #44. @jzaefferer please review.

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,47 +3,20 @@
   "version": "1.0.0",
   "scripts": {
     "build": "rollup -c",
-    "lint": "eslint lib/ test/",
+    "lint": "standard",
     "test": "npm run lint && node test/test.js && npm run test-integration",
     "test-integration": "mocha test/integration/adapters.js"
   },
   "devDependencies": {
-    "babel-eslint": "^5.0.0",
     "babel-preset-es2015-rollup": "^1.1.1",
     "chai": "^3.5.0",
     "commitplease": "^2.2.3",
-    "eslint": "~2.2.0",
     "jasmine": "^2.3.1",
     "jasmine-core": "^2.3.4",
     "mocha": "^2.4.5",
     "qunitjs": "^1.18.0",
     "rollup": "^0.26.0",
-    "rollup-plugin-babel": "^2.4.0"
-  },
-  "eslintConfig": {
-    "parser": "babel-eslint",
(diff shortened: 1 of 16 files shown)


jzaefferer:
👍 

Nice work. Did you do this all by hand, or use `standard-format`?

Somewhat a note to myself: For the future, I'd try to group the various changes: Removes semicolons in one commit, switch quotes in another, finally change whitespace (or a different order). That way the whitespace changes could be verified with `-w` (`?w=1` on GitHub), which should result in an empty diff. Mixing it all together makes it harder.

package.json
```diff
@@ -3,47 +3,20 @@
   "version": "1.0.0",
   "scripts": {
     "build": "rollup -c",
-    "lint": "eslint lib/ test/",
+    "lint": "standard",
```
jzaefferer: We don't need a separate script for this anymore. Let's inline it into the `test` script.

jzaefferer: Addressed that in b0dc0955a381b74d4aed54031548105078fd2365

flore77:
Yes, I have used `standard-format`, it would have been insane to do it all by hand, there where approximately **750 errors**. After using `standard-format`, there were only 10 or something like that.

> I'd try to group the various changes:

Yes it would be good, but I don't know if you can achieve this with `standard-format`.

jzaefferer:
> After using standard-format, there were only 10 or something like that.

Which one did it miss?

> but I don't know if you can achieve this with standard-format.

Probably not out of the box. Not sure if its worth investigating...

flore77:
> Which one did it miss?

He missed to remove some blank lines, like:
```js
for () {

// comment
code
}
```

There should be no line btw for and the comment.

Variables declared:
```js
var x, y, z;
```
that must be declared:
```js
var x;
var y;
var z;
```

And there were some errors about not knowing who `it` or `describe` was, one line at the top telling `standard`that there is a `Mocha env`, fixed them all.

I think that was all.

> Probably not out of the box. Not sure if its worth investigating...

Yes, I think there is no gain to investigate that, for this project.

PR #46, opened 2016-04-26
Mocha adapter

flore77:


test/mocha/index.html
```diff
@@ -0,0 +1,24 @@
...
+    <script>mocha.setup('bdd')</script>
+    <script src="../../dist/js-reporters.js"></script>
+    <script src="tests.js"></script>
+    <script>
+        window.onload = function () {
+            mocha.run();
+
+            var runner = new JsReporters.MochaAdapter(mocha);
+            JsReporters.ConsoleReporter.init(runner);
+            mocha.run();
```
jzaefferer: Why is this 2nd `mocha.run()` call necessary?

flore77:
The tests are passing so it should be good :smiley: 

It is ok to call Mocha **twice**, because on first run, Mocha uses it's default html reporter, and then when we are passing it to our adapter, it is overriding the html reporter.

Now I was checking for race conditions, it seems mocha runs them in parallel, so it should be fine.

jzaefferer:
Can you add a comment to the code, to include that explanation?

flore77:
Sure, I was thinking about that too.

 And what I do mean through runs them in parallel, is that mocha runs  them **independently** in parallel, with help of different [processes](https://github.com/mochajs/mocha/blob/c0f9be244bff479e948f59d1bdb825a45c2cb40c/bin/mocha#L57). 
But now this is happening when used from command line, I'm not sure, when it is used programmatic. @jzaefferer  Should I investigate a little more ?

jzaefferer:
I think we can land this and investigate that separately. Would be nice to be able to add a reporter, instead of replacing the existing one.

Is there anything else left to do here? Up to date with master?

jzaefferer:
Now that #41 landed, can you rebase and update this, to make use of the external events module as well?

flore77:
@jzaefferer done.

jzaefferer:
Two more things:

1. Let's add `npm run build` to the test script, otherwise we may run tests against an outdated built file.
1. Extend the integration tests with the mocha adapter

flore77:
@jzaefferer finished. I add a `pretest` script, which calls the `building script` before actual testing.

test/integration/adapters.js
```diff
@@ -15,7 +15,7 @@ var collectedData = {}
  * @type {Array}
  * @const
  */
-const adapters = ['Jasmine', 'QUnit']
+const adapters = ['Jasmine', 'QUnit', 'Mocha']
```
jzaefferer: Let's replace this with `Object.keys(runAdapters)` (or something like that)

lib/adapters/MochaAdapter.js
```diff
@@ -0,0 +1,112 @@
+import EventEmitter from 'events'
+import { Test, Suite } from '../Data.js'
+
+export default class MochaAdapter extends EventEmitter {
+  constructor (mocha) {
+    super()
+
+    this.mocha = mocha
+    mocha.reporter((runner) => {
```
jzaefferer: Is this documented to replace the current reporter? Is there an option to add an additional reporter instead?

lib/index.js
```diff
@@ -8,6 +9,7 @@ import { Test, Suite } from './Data.js'
 export default {
   QUnitAdapter: QUnitAdapter,
   JasmineAdapter: JasmineAdapter,
+  MochaAdapter: MochaAdapter,
```
jzaefferer: Let's use object shorthand, omitting the value when it matches the key.

test/test.js
```diff
@@ -25,11 +27,27 @@ var qunitRunner = new JsReporters.QUnitAdapter(QUnit)
...
 require('./qunit/tests.js')
 
 QUnit.load()
+
+var mocha = new Mocha()
+var mochaRunner = new JsReporters.MochaAdapter(mocha)
+var mochaTestReporter = new JsReporters.TestReporter(mochaRunner, referenceData.Mocha)
+
+mocha.addFile(
+  path.join(__dirname, 'mocha/tests.js')
```
jzaefferer: Why is this `addFile` call using `path.join`, but the other one (above) is just adding `test/mocha/tests.js`?

lib/adapters/MochaAdapter.js
flore77: [Mocha Usage](https://mochajs.org/#usage), the --reporter description is `specify the reporter to use` and [Using Mocha programmatically](https://github.com/mochajs/mocha/wiki/Using-mocha-programmatically#set-options), `reporter` function description is `set the reporter`. 

So there is no option like `addReporter` or something like that :disappointed:

test/test.js
flore77: Which one above? Do you mean this [one](https://github.com/js-reporters/js-reporters/pull/46/files?diff=unified#diff-cbe814bd39e9f048a8d19ed4d2006bd5R46). I think it can take also `asolute paths` as also `relative to the current directory`.

test/integration/adapters.js
flore77: Sure. But I think we should change this in another issue/branch, where we are refactoring testing, because there will be one, for adding more tests, use the same `referenceData obj`.

lib/index.js
flore77: Ok. May I move this file to the `root` of the project ? I think there should be the `entry point`.

Maybe also doing this in another issue, where we do more small fixes ?

flore77:
I used absolute paths for `Mocha` and `QUnit`, but for `Jasmine` it does not work.

flore77:
How strange, I don't know how [fe37573](/js-reporters/js-reporters/pull/46/commits/fe375739c966e6f2f880722055054838a16f3c5b) has passed, there should have been `2 linting errors`.

jzaefferer:
Alright, landed! I did the `Object.keys` fix, along with a cleanup of some source comments, in 255209e8011bb7bddbd38da19fcf64bdb6f8f7ad

PR #48, opened 2016-04-29
Entry point

flore77:
@jzaefferer please review.


jzaefferer:
Could you try to redo that 2nd commit, using `git mv`, so that it recognises the rename?

flore77:
I had to broke it, in two commits, because when I was modifying the `renamed index.js` file, than `git` was seeing it like a `new file` and the` old one deleted`.

jzaefferer:
Can you rebase it to get rid of the revert commit?

flore77:
@jzaefferer done.

jzaefferer:
The commit message for the first commit is a bit messy. Normally I'd just fix this while merging, in this case I think its worth if you do it yourself and pay a little more attention to those in the future ;-)

flore77:
Yes, sorry about that.

jzaefferer:
Alright, that's done.

Issue #47, opened 2016-04-28
Update QUnit version and adapter to make use of nested modules support


jzaefferer:
Newer QUnit versions support nested modules. Maybe that works well enough that we can use the same `test/referenceData.js` for all 3 adapters.

Would also be good to test QUnit 2.0.0-rc1 (or newer), to make sure our adapter works with that.

jzaefferer:
@flore77 I think this would be a good next step. Should help bring the tests in line.

flore77:
Sure :+1:

flore77:
Solved in #58 and #60.

PR #49, opened 2016-05-09
One reference data object for all adapters

flore77:
This should solve #47 .

I have built a new Mocha test fixture, as also I wrote some tests for it and some bugs are already emerging (some tests are failing).

I would have also add a **global test**, a test with is not wrapped into a suite, this test would arrive into the global suite, Mocha and Qunit permit this, but **Jasmine no**, so I didn't included it, I will think to something to add it in the future.

I'm waiting your feedback @jzaefferer, to know if I can go further. Thanks!

### DIFF
--- a/test/integration/adapters.js
+++ b/test/integration/adapters.js
@@ -1,15 +1,14 @@
 /* eslint-env mocha */
 
 var expect = require('chai').expect
-var data = require('../referenceData')
+var refData = require('./reference-data.js')
 var runAdapters = require('./adapters-run.js')
 
 // Collecting the adapter's output.
-var collectedData = {}
+var collectedData
 
 function _collectOutput (eventName, done, eventData) {
-  // Assume now (for simplicity) that there is only one event per type.
-  collectedData[eventName] = eventData
+  collectedData.push([eventName, eventData])
   done()
 }
 
@@ -30,20 +29,19 @@ function _attachListeners (done, runner) {
 describe('Adapters integration', function () {
   Object.keys(runAdapters).forEach(function (adapter) {
     describe(adapter + ' adapter', function () {
-      var referenceData = data[adapter]
+      var testDescription
 
       before(function (done) {
-        collectedData = {}
+        collectedData = []
         runAdapters[adapter](_attachListeners.bind(null, done))
       })
 
-      describe('Global suite', function () {
-        it('should have no name', function () {
-          expect(referenceData[0][1].name).to.be.equal(collectedData.runStart.name)
-        })
+      refData.forEach(function (value, index) {
+        testDescription = value[2]
 
-        it('should have all the other suites as childSuites', function () {
-          expect(referenceData[0][1].childSuites).to.have.lengthOf(collectedData.runStart.childSuites.length)
+        it(testDescription, function () {
+          expect(collectedData[index][0]).equal(value[0])
+          expect(collectedData[index][1]).to.be.deep.equal(value[1])
         })
       })
     })
--- a/test/integration/adapters.js
+++ b/test/integration/adapters.js
@@ -28,7 +28,7 @@ function _attachListeners (done, runner) {
 
 describe('Adapters integration', function () {
   Object.keys(runAdapters).forEach(function (adapter) {
-    describe(adapter + ' adapter', function () {
+    describe.skip(adapter + ' adapter', function () {
       var testDescription
 
       before(function (done) {
--- a/package.json
+++ b/package.json
@@ -15,7 +15,7 @@
     "jasmine": "^2.3.1",
     "jasmine-core": "^2.3.4",
     "mocha": "^2.4.5",
-    "qunitjs": "^1.18.0",
+    "qunitjs": "^2.0.0-rc1",
     "rollup": "^0.26.0",
     "rollup-plugin-babel": "^2.4.0",
     "rollup-plugin-commonjs": "^2.2.1",
--- a/test/fixtures/jasmine.js
+++ b/test/fixtures/jasmine.js
@@ -1,0 +1,49 @@
+/* global describe, it, xit */
+
+it('global test', function () {
+
+})
+
+describe('suite with passing test', function () {
+  it('should pass', function () {
+
+  })
+})
+
+describe('suite with skipped test', function () {
+  xit('should skip', function () {
+
+  })
+})
+
+describe('suite with failing test', function () {
+  it('should fail', function () {
+    throw new Error('error')
+  })
+})
+
+describe('suite with tests', function () {
+  it('should pass', function () {
+
+  })
+
+  xit('should skip', function () {
+
+  })
+
+  it('should fail', function () {
+    throw new Error('error')
+  })
+})
+
+describe('outter suite', function () {
+  describe('inner suite', function () {
+    it('inner test', function () {
+
+    })
+  })
+
+  it('outter test', function () {
+
+  })
+})
--- a/test/fixtures/mocha.js
+++ b/test/fixtures/mocha.js
@@ -1,0 +1,49 @@
+/* global describe, it */
+
+it('global test', function () {
+
+})
+
+describe('suite with passing test', function () {
+  it('should pass', function () {
+
+  })
+})
+
+describe('suite with skipped test', function () {
+  it.skip('should skip', function () {
+
+  })
+})
+
+describe('suite with failing test', function () {
+  it('should fail', function () {
+    throw new Error('error')
+  })
+})
+
+describe('suite with tests', function () {
+  it('should pass', function () {
+
+  })
+
+  it.skip('should skip', function () {
+
+  })
+
+  it('should fail', function () {
+    throw new Error('error')
+  })
+})
+
+describe('outter suite', function () {
+  describe('inner suite', function () {
+    it('inner test', function () {
+
+    })
+  })
+
+  it('outter test', function () {
+
+  })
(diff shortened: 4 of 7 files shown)


jzaefferer:
Looking fine so far. Let me know when you've made some more progress. Currently there are too many failures for me to tell which one are expected and which one need discussion.

flore77:
I have done a better research and added also a **global test**, because they are also permitted in an **Jasmine env**. 

There are **2 errors** so far, there are multiple failures regarding the same error.

Ok, I will finish writing the whole `test set` and then I will make a detailed list with the errors that have emerged.

flore77:
@jzaefferer  I finished writing the tests, they are a little bit massive, maybe we can reduce them. I was thinking to let only `one suite with one test` and to remove the others 2. I am open to suggestions.

The following errors have emerged:

1. The objects passed on `testStart` event also have the props `runtime, status, errors`, which are `undefined`. Is it ok? Or the object should have only the props: `testName` and `suiteName`. Another issue is that on the `testEnd` event is passed the **same object as on testStart**, so the object is mutated and we end up in our `collectedData` array, to already know, the status, runtime etc. of the test, which is only starting. Shouldn't we pass another object on the `testEnd` event ? 

2. This is really a bug: if we have `multiple tests` in `different suites` with the `same name`, then 
all this tests will end up with the `suiteName` of the last suite, even they belong to other suites.

3. For the nested suites, the `inner test` has as `suiteName` the `concatenated names` of the `outter suite` and the `inner suite` is this ok? Or it should have only the name of the `inner suite` ?

Also I was thinking, for the nested suites, the outter suite contains a test and another suite, the inner suite, the questions would be: which to run first, the test or the inner suite ? 
Mocha runs first the tests and then the inner suites, I have not yet test for the others. I hope that the others are doing it also this way, so we can accept it as a `defacto standard`.

jzaefferer:
> I finished writing the tests, they are a little bit massive, maybe we can reduce them. I was thinking to let only one suite with one test and to remove the others 2. I am open to suggestions.

I'll take a look. Though I tend to ignoring this until we figure out the overall structure. Make it correct first, refactor later.

> The objects passed on testStart event also have the props runtime, status, errors, which are undefined. Is it ok? Or the object should have only the props: testName and suiteName. 

As long as the values are `undefined` its probably okay, but we can get back to that. I suspect it might be resolved by other changes.

> Another issue is that on the testEnd event is passed the same object as on testStart, so the object is mutated and we end up in our collectedData array, to already know, the status, runtime etc. of the test, which is only starting. Shouldn't we pass another object on the testEnd event ?

I think that's wrong, we shouldn't mutate any objects that we emit.

> This is really a bug: if we have multiple tests in different suites with the same name, then 
all this tests will end up with the suiteName of the last suite, even they belong to other suites.

Agreed, that needs to be fixed.

> For the nested suites, the inner test has as suiteName the concatenated names of the outter suite and the inner suite is this ok? Or it should have only the name of the inner suite ?

I'm not sure about that. May need to go back and check what the various testing frameworks are doing here, adjust the spec etc. Probably better to file a separate issue for that and ignore/work around that here.

Overall this is very promising. Nice to see tests uncovering bugs :-)

test/integration/adapters.js
```diff
@@ -27,23 +26,258 @@ function _attachListeners (done, runner) {
...
+  it('should be emitted on suiteStart event', function () {
+    expect(collectedData[index][0]).to.be.equal('suiteStart')
+  })
+
+  it('should have a name', function () {
+    expect(collectedData[index][1].suiteName).to.be
+      .equal(refData[index][1].suiteName)
+  })
+
+  switch (type) {
```
jzaefferer: This is not a good idea. At least use something readable ("simple" instead of "s"), better yet, pass functions around. So instead of `_testSuite(3, 's')`, do something like `_testSuite(3, _testSimpleSuite)`, then call that function with the `index` argument.

jzaefferer:
I wonder if we can simplify the tests by iterating over `collectedData`, and comparing each entry with the equivalent entry in the `referenceData` array (`deep.equal`). We can add messages for each `expect` based on the names or other properties, to help with debugging.

flore77:
I don't know, it would be very hard to add messages and to define a good structure with `describes` and `its`. I will try to achieve something.

But definitely I must do something with the numbers (ex. `testSuite(3), testSuite(20)`), it is very annoying and if the structure changes a little bit, than you have to update the whole tests. 

I have tried something with an `index`, that was incremented in each `describe`, but `Mocha` is doing a parsing of the tests, and takes the call of the functions in the beginning with the value of the parameters at that moment...and then the increment is useless...

test/integration/adapters.js
```diff
@@ -192,7 +171,7 @@ describe('Adapters integration', function () {
       })
 
       describe('Suite with only one passing test', function () {
-        _testSuite(3, 's')
+        _testSuite(3, _testSimpleSuite.bind(null))
```
jzaefferer: Why do you need the `bind` call? If you really need a specific context, use `call` or `apply` at the call site, like `testSpecificSuite.call(null, index)`

jzaefferer:
> I don't know, it would be very hard to add messages and to define a good structure with describes and its. I will try to achieve something.

In case the idea wasn't well enough explained, I'm having something like this in mind:

```
describe('...', () => {
  referenceData.forEach(({ eventName, eventData}, index) => {
    it(`Event ${index} should be ${eventName} with the right data`, () => {
      expect(referenceData[index][0]).equal(eventName)
      expect(referenceData[index][1]).equal(eventData)
    })
  })
})
```
The quality of your tests then mostly depends on the quality of your reference data. Changes in the tests require changes in the reference data, but, among other advantages, you don't need to keep track of indexes.

If this looses too much precision, you could start with this, then test specific edge cases, instead of having to write individual tests for everything.

test/integration/adapters.js
flore77: No, they don't need to run in a specific context, I removed the binding.

flore77:
Yeah, I cannot decide. I also wrote the above version, I took some time to analyse, which would be the best version, but the purpose of the tests is not clear, when you see something failing, you don't know what has really failed, you must look carefully to the test output. 

Although, it is a really nice and simple way to write tests, and how you have said, now test changes would require only changes in the `refData file`.

I have adopt the above implementation as also added a new field into the `refData` array, a description, which is usefull for test displaying, which describes perfectly the flow of the `test fixture`, so if a test fails, you will now which part of the test fixture caused the problem and where to look into the `refData file`. What do you think @jzaefferer ?

flore77:
I have updated the `new-ref-data file` to match the new testing style. Now I wanted to make some fixes to the `Mocha Adapter`.

- One fix is to allow `multiple suites/tests with the same name`, to achieve this I have put on Mochas suite/test objects an `index`, this way each suite/test will have it's `own key`. Before, the key was the suite/test `title`, this way multiple objects could have the same key. I think this is the simplest solution.

- The second fix should be that the objects that we emit, should then be not mutated. To achieve this I was thinking to do some deep copies, and to emit those. I tried to achieve this with some exiting packages, but they do not make a really deep copy, for example for the `suite objects`, that contain other `test objects`, they only copy the reference to it... It is good to modify the objects, because on the `testEnd`, or `suiteEnd` events we need to emit some updated objects, another solutions would be to create other objects for the end events and emit those.

@jzaefferer what would you suggest?

lib/adapters/MochaAdapter.js
```diff
@@ -16,48 +20,51 @@ export default class MochaAdapter extends EventEmitter {
...
-    this.idDataMap = {}
   }
 
   convertSuite (mochaSuite) {
     var name = mochaSuite.title
     var childSuites = []
     var tests = []
 
+    mochaSuite.index = this.index
+    this.index++
```
jzaefferer: I strongly prefer `this.index += 1`, makes the change for explicit and discourages inline increments (change var and use it at the same time)

jzaefferer:
I'm concerned about the scope of this PR, I think its getting out of hand. There are a lot of style changes, test updates, var renames and other refactorings and other changes mixed together. Tests aren't passing, so I'm sure this can't yet be merged. Reviewing many changed files over and over is bad.

For a start, the changes in `test/qunit/tests.js` really don't belong here, same for the equivalent files for the other frameworks. Let's split those out and merge them separately. 

Next we should focus on landing the new approach for testing, where we skip tests that cause failures due to bugs in the reporters. We should still run all the tests that pass (including Jasmine and QUnit). 

We can then unskip those tests while fixing bugs in the adapters. We can do those separately from variable and argument renames (like renaming the `details` argument to `mochaTest` etc.).

Generally let's not version changes (test/integration/adapters.js vs test/integration/new-adapters.js). We can easily roll back by reverting (or deleting) commits. There's no need to make copies.

flore77:
@jzaefferer I refactored the branch.

I don't know how to land this without solving bugs in the adapters, there are massive errors. For example for `QUnit` every single test is failing, because they are shifted, the `globalSuite` is emitted on `runStart` event, but also on `suiteStart` event...

How should we approach this ?

jzaefferer:
As-is this looks good, except for the test failures. It looks like all three adapters fail, so maybe we can just do a `describe.skip` on the entire suite for now, merge, then revert that while fixing failures step by step. What do you think?

Btw. Mocha's `bail` option might be worth a try, helps to focus on a single failing test at a time.

flore77:
Sure, now this should be good to merge.

PR #51, opened 2016-05-21
Fix mocha adapter

### LINKED ISSUE #50: Fix Mocha Adapter

## Issues

[1] The current adapter do not permit `suites` or `tests` with the same name, for example:

```js
describe('suite1', function() {
  it('test1', function() {})
})

describe('suite2', function() {
  it('test1', function() {})
})
```
Both tests will have the same suite as suite parent, which is `suite2`.

This is happening because the `tests` are stored in an [object](https://github.com/js-reporters/js-reporters/blob/9aba15bc8a801b5a1b8f0829b7fb534cfa46317a/lib/adapters/MochaAdapter.js#L60) by their `title` (description) which is non-unique, the same happens for `suites`.

[2] The objects that are **emitted** are then **mutated**. [Emitted](https://github.com/js-reporters/js-reporters/blob/9aba15bc8a801b5a1b8f0829b7fb534cfa46317a/lib/adapters/MochaAdapter.js#L81) and [mutated](https://github.com/js-reporters/js-reporters/blob/9aba15bc8a801b5a1b8f0829b7fb534cfa46317a/lib/adapters/MochaAdapter.js#L89-L97).

[3] The `inner suite`, in `nested suites`, has as name the concatenation of the `names` from the `top outter suite` to the `most inner suite`. This is done by us in [our adapter](https://github.com/js-reporters/js-reporters/blob/9aba15bc8a801b5a1b8f0829b7fb534cfa46317a/lib/adapters/MochaAdapter.js#L47-L54).

[4]  The `tests` on the `startTest` event, have also the properties `runtime`, `status`, `errors`, they are undefined, but maybe they shouldn't be there at all.

## Solutions

[1] See [2] solves also this.

[2] Construct every time new objects, by calling the convert functions, without indexing the objects. When converting a `test` check if the `Mocha object` has an `error` property, if it has then we must build a `test end`, else a `test start`.

[3] Concatenate the suites names. 

[4] Leave it like this, with the props to be undefined.

flore77:
Fix #50.

### DIFF
--- a/lib/adapters/MochaAdapter.js
+++ b/lib/adapters/MochaAdapter.js
@@ -107,6 +107,9 @@ export default class MochaAdapter extends EventEmitter {
   }
 
   onEnd () {
-    this.emit('runEnd', this.convertSuite(this.globalSuiteEnd))
+    var suite = this.convertSuite(this.globalSuiteEnd)
+    suite.name = undefined
+
+    this.emit('runEnd', suite)
   }
 }
--- a/lib/adapters/MochaAdapter.js
+++ b/lib/adapters/MochaAdapter.js
@@ -16,7 +16,6 @@ export default class MochaAdapter extends EventEmitter {
       runner.on('suite end', this.onSuiteEnd.bind(this))
       runner.on('end', this.onEnd.bind(this))
     })
-    this.idDataMap = {}
   }
 
   convertSuite (mochaSuite) {
@@ -35,7 +34,7 @@ export default class MochaAdapter extends EventEmitter {
       childSuites,
       tests
     )
-    this.idDataMap[mochaSuite.title] = suite
+
     return suite
   }
 
@@ -57,28 +56,26 @@ export default class MochaAdapter extends EventEmitter {
       testName,
       suiteName
     )
-    this.idDataMap[mochaTest.title] = test
+
     return test
   }
 
   onStart () {
     var startSuite = this.convertSuite(this.runner.suite)
     startSuite.name = undefined
-    this.idDataMap.globalSuite = startSuite
 
     this.emit('runStart', startSuite)
   }
 
   onSuite (details) {
-    var suite = this.idDataMap[details.title]
-    if (suite !== this.idDataMap.globalSuite) {
-      this.emit('suiteStart', suite)
+    if (!details.root) {
+      this.emit('suiteStart', this.convertSuite(details))
     }
   }
 
   onTest (details) {
     this.errors = []
-    this.emit('testStart', this.idDataMap[details.title])
+    this.emit('testStart', this.convertTest(details))
   }
 
   onFail (test, error) {
@@ -86,7 +83,7 @@ export default class MochaAdapter extends EventEmitter {
   }
 
   onTestEnd (details) {
-    var test = this.idDataMap[details.title]
+    var test = this.convertTest(details)
     if (details.pending) {
       this.emit('testStart', test)
       test.status = 'skipped'
@@ -100,13 +97,15 @@ export default class MochaAdapter extends EventEmitter {
   }
 
   onSuiteEnd (details) {
-    var suite = this.idDataMap[details.title]
-    if (suite !== this.idDataMap.globalSuite) {
-      this.emit('suiteEnd', suite)
+    if (!details.root) {
+      this.emit('suiteEnd', this.convertSuite(details))
+    }
+    else {
+      this.globalSuiteEnd = details
     }
   }
 
   onEnd () {
-    this.emit('runEnd', this.idDataMap.globalSuite)
+    this.emit('runEnd', this.convertSuite(this.globalSuiteEnd))
   }
 }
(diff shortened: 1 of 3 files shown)


flore77:
@jzaefferer please review this, I solved [1] and [2].

I put an `errors` property on `Mocha's test obj`, because the errors that we are collecting, are not on the `test obj`, they are passed as a parameter on `Mocha's fail event`. So with this prop I saved the errors, but also we can identify if we want to convert a `testStart` or a `testEnd`.

Mocha has a property `err` on the tests that failed, which should be the same error passed on the `fail event`, but I don't know why the objects that Mocha passes to our adapter do not have this prop, with this prop we could remove the `errors` prop added by us. I will investigate more.

lib/adapters/MochaAdapter.js
```diff
@@ -53,60 +52,69 @@ export default class MochaAdapter extends EventEmitter {
       parent = parent.parent
     }
 
-    var test = new Test(
-      testName,
-      suiteName
-    )
-    this.idDataMap[mochaTest.title] = test
+    var test
```
jzaefferer: I'd remove this and the `return test` later, and do 2 returns from the `if/else`.

jzaefferer: Actually a return from the `if`, then the else won't be needed anymore and be just `return new Test...`

lib/adapters/MochaAdapter.js
```diff
@@ -53,60 +52,69 @@ export default class MochaAdapter extends EventEmitter {
...
     this.emit('testEnd', test)
   }
 
   onSuiteEnd (details) {
-    var suite = this.idDataMap[details.title]
-    if (suite !== this.idDataMap.globalSuite) {
-      this.emit('suiteEnd', suite)
+    if (!details.root) {
+      this.emit('suiteEnd', this.convertSuite(details))
+    }
```
jzaefferer: Remove linebreak, should be `} else {`. Also verify why linter isn't checking this.

jzaefferer: Nevermind, it did check it, you just ignored the linter /o\

jzaefferer:
There's still (silly) linting issues, which prevent the actual tests from running on Travis.

flore77:
:tada: tests are passing. 

But we still have an issue :disappointed: We must put `tests runtime` on `0` also in the `suite objects` emitted on the `suiteEnd` event. Doing this here or on another branch?

flore77:
I have found out, what is happening with Mocha `err` property :tada: 

The `err` property is not part of the `test object`, it is added by Mocha [Base reporter](https://github.com/mochajs/mocha/blob/015b809c27109d175f0e51af1b58b0c0310261b2/lib/reporters/base.js#L279), this is why we did not find it in our reporter. Basically it is the same thing that we are doing with our `error` property.

Should we change our property also to `err` to be more appropriate to Mocha's style ?

jzaefferer:
Its strange that the reporter would modify the incoming data. That seems like something that should happen inside the regular framework, not in a reporter.

I think we're fine with adding the `errors` property like we currently do.

flore77:
@jzaefferer good to merge?

lib/adapters/MochaAdapter.js
```diff
@@ -53,60 +52,64 @@ export default class MochaAdapter extends EventEmitter {
...
+      test = this.convertTest(details)
       this.emit('testStart', test)
-      test.status = 'skipped'
-    } else {
-      test.status = details.state
     }
-    test.runtime = details.duration
-    test.errors = this.errors
+
+    details.errors = this.errors
```
jzaefferer: Why are we adding them instead of checking for `this.errors` inside of `convertTest`?

flore77: We are adding it, because we need to stores the `errors` so that when we are emitting the suites on the suiteEnd event, the tests also contains error, status and runtime.

jzaefferer: Right, but we don't need to add it to the Mocha object. You could check `this.errors` inside `convertTest` instead. Maybe do a lazy init inside `onFail`, something like this:

```
onFail (test, error) {
  if (!this.errors) {
    this.errors = []
  }
  this.errors.push(error)
}
```

Then in `convertTest`, you can check for `this.errors`.

flore77: But on the `suiteEnd` event when we receive a Mocha suite and this suite contains multiple tests. When we will convert it, and implicit we will convert the tests too, how we will know the `error` for each test? `this.errors` maybe would contain the error from the last test, but thats all.

flore77: Also what dou you mean with the lazy init:

```js
if (!this.errors) {
  this.errors = []
}
```

When should this execute ? Only one time, with the first test ? And to keep all errors from all tests inside this array ? I don't get the idea :disappointed:

jzaefferer: Nevermind, let's leave it as it is right now, but add a fat exclaimer comment explaining why we're modifying the Mocha object.

flore77:
@jzaefferer done.

jzaefferer:
Excellent! Good to merge.

Which adapter are you going to do next? I suggest QUnit, since 2.0.0 isn't final and we might be able to give them some valuable feedback.

flore77:
Sure.

PR #52, opened 2016-05-21
Fix ref tests/suites names

flore77:
Should fix [3] from #50.


flore77:
@jzaefferer  I would like to take care on this pr also of the `tests runtime`. In the ref data file the runtime is 0, Mocha runs them in 0 milliseconds, but it can happen that it will run them in 1 milliseconds (I think that it cannot take more than this, because there are simple tests, some of them also void).

As a solution, I think to write one test that checks that all `test's run times` are in the range `0 - 1 milliseconds`, and then put all run times on 0 for the deep equal check.

jzaefferer:
👍 

You can merge this to master yourself.

PR #53, opened 2016-05-21
Standardize tests run time to match with the reference data

flore77:
@jzaefferer I would like to take care on this pr of the `tests runtime`. In the ref data file the runtime is 0, Mocha runs them in 0 milliseconds, but it can happen that it will run them in 1 milliseconds (I think that it cannot take more than this, because there are simple tests, some of them also void). As a solution, I think to write one test that checks that all `test's run times` are in the range `0 - 1 milliseconds`, and then put all run times on 0 for the deep equal check.

### DIFF
--- a/test/integration/adapters.js
+++ b/test/integration/adapters.js
@@ -36,10 +36,24 @@ describe('Adapters integration', function () {
         runAdapters[adapter](_attachListeners.bind(null, done))
       })
 
+      it('tests runtime should be between 0 and 1 ms', function () {
+        collectedData.forEach(function (value) {
+          if (value[0] === 'testEnd' && value[1].status !== 'skipped') {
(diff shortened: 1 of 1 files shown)


test/integration/adapters.js
```diff
@@ -36,10 +36,23 @@ describe('Adapters integration', function () {
...
+        })
+      })
+
       refData.forEach(function (value, index) {
         testDescription = value[2]
 
         it(testDescription, function () {
+          if (collectedData[index][0] === 'testEnd' &&
+              collectedData[index][1].status !== 'skipped') {
+            collectedData[index][1].runtime = 0
```
jzaefferer: Add a comment here, explaining why we're changing this.

flore77:
@jzaefferer done.

jzaefferer:
👍

PR #54, opened 2016-05-22
 Set suite tests runtime also to 0ms.

flore77:
This fixes tests runtime in the suite objects emitted on the `suiteEnd` event.


flore77:
@jzaefferer please review this.

jzaefferer:
Looks good. Let's land this, then rebase #51.

PR #58, opened 2016-05-30
Fix qunit adapter

### LINKED ISSUE #57: Fix QUnit Adapter

## Issues

[1] The `QUnit global module` is emitted also on [suiteStart](https://github.com/js-reporters/js-reporters/blob/49b51f43ce8ed2d1b05a95def7250883ad97c7b9/lib/adapters/QUnitAdapter.js#L46) event, as also on [suiteEnd](https://github.com/js-reporters/js-reporters/blob/49b51f43ce8ed2d1b05a95def7250883ad97c7b9/lib/adapters/QUnitAdapter.js#L78).

[2] Passing tests are failing, because QUnit is checking for an `assertion` and in our [test fixture](https://github.com/js-reporters/js-reporters/blob/49b51f43ce8ed2d1b05a95def7250883ad97c7b9/test/fixtures/qunit.js) there are none.

[3] Objects are **emitted** are then **mutated**. [Emitted](https://github.com/js-reporters/js-reporters/blob/49b51f43ce8ed2d1b05a95def7250883ad97c7b9/lib/adapters/QUnitAdapter.js#L51) and [mutated](https://github.com/js-reporters/js-reporters/blob/49b51f43ce8ed2d1b05a95def7250883ad97c7b9/lib/adapters/QUnitAdapter.js#L63-L72).

[4] The current adapter do not permit `suites` with the [same name](https://github.com/js-reporters/js-reporters/blob/49b51f43ce8ed2d1b05a95def7250883ad97c7b9/lib/adapters/QUnitAdapter.js#L29).

[5] Skipped tests have as `runtime 0ms`.

[6] Nested modules run. QUnit runs the nested modules (suites) from the inner most suite to the outter most suite.

[7] QUnit provide already for nested modules a concatenated name. The suite names are separated by the character ">", ex: `outter suite > inner suite`.

[8] Also for failing tests, QUnit is taking the error and creates an object with some more properties, so the deep equal assertion will fail.

[9] The `globalSuite` emitted on `suiteStart` is not created properly.

## Solutions

- [x] [1] - Check on the `suiteStart` and `suiteEnd` event, that the received module is not the `globalModule` and only then emit a converted suite.

- [x] [2] We can add a dummyAssertion in our fixture, like: `ok(true)`.

- [x] [3], [4] - We should convert the `modules` and `tests` every time we are emitting something, the same as in the [MochaAdapter](https://github.com/js-reporters/js-reporters/blob/49b51f43ce8ed2d1b05a95def7250883ad97c7b9/lib/adapters/MochaAdapter.js)

- [x] [5] I think it would be a good thing for QUnit to also put the `runtime` of a  `skipped test` to be `undefined`, because that test is no running, so it doesn't make sens to have a runtime. For now we can set it to undefined [by hand](https://github.com/js-reporters/js-reporters/blob/49b51f43ce8ed2d1b05a95def7250883ad97c7b9/lib/adapters/QUnitAdapter.js#L71) in our adapter.

- [x] [6] We will run it like Mocha, from the outer most to the inner most, also because the QUnit reporter displays tests this way.

- [x] [7] Remove the ">" character.

- [x] [8] We can do something like with the [runtimes](https://github.com/js-reporters/js-reporters/blob/49b51f43ce8ed2d1b05a95def7250883ad97c7b9/test/integration/adapters.js#L52-L58), to make a function that checks that the errors are conform with the QUnit errors and then override them.

- [x] [9] Create a function that builds the global suite for the runStart and runEnd events.

flore77:
Fixes #57.

### DIFF
--- a/lib/adapters/QUnitAdapter.js
+++ b/lib/adapters/QUnitAdapter.js
@@ -31,11 +31,9 @@ export default class QUnitAdapter extends EventEmitter {
    * @param suiteNameParam - is used only when calling convertTest from
    *  the convertModule function, in rest it is undefined.
    */
-  convertTest (qunitTest, suiteNameParam) {
-    var suiteName = suiteNameParam
-
-    // Assign the name of the module, only if it is not the global one, then
-    // leave the suiteName to be undefined.
+  convertTest (qunitTest, suiteName) {
+    // Set the suiteName only if it is undefined and it is not the name of the
+    // globalModule.
     if (!suiteName && qunitTest.module !== '') {
       suiteName = qunitTest.module
     }
--- a/lib/adapters/QUnitAdapter.js
+++ b/lib/adapters/QUnitAdapter.js
@@ -53,7 +53,6 @@ export default class QUnitAdapter extends EventEmitter {
 
       return new Test(qunitTest.name, suiteName, status, qunitTest.runtime,
           qunitTest.errors)
-
     }
 
     return new Test(qunitTest.name, suiteName)
--- a/lib/adapters/QUnitAdapter.js
+++ b/lib/adapters/QUnitAdapter.js
@@ -14,14 +14,26 @@ export default class QUnitAdapter extends EventEmitter {
     QUnit.done(this.onDone.bind(this))
 
     this.testsDetails = {}
+    this.emitGlobal = false
   }
 
   convertModule (qunitModule) {
-    var suiteName = qunitModule.name
+    // The Test needs the concatenated name of the suite only without ">".
+    var suiteName = qunitModule.name.replace(/> /g, '')
     var tests = []
 
     for (let qunitTest of qunitModule.tests) {
       tests.push(this.convertTest(qunitTest, suiteName))
+    }
+
+    // If the suite is for the "createGlobalSuite" function, then the name must
+    // be the name provided by QUnit, else keep only the suite name itself.
+    if (this.emitGlobal) {
+      suiteName = qunitModule.name
+    } else {
+      var index = qunitModule.name.lastIndexOf('>')
+
+      suiteName = (index !== -1) ? suiteName.substring(index + 2) : suiteName
     }
 
     return new Suite(suiteName, [], tests)
@@ -85,6 +97,8 @@ export default class QUnitAdapter extends EventEmitter {
     var globalSuite
     var indexStart, indexEnd
 
+    this.emitGlobal = true
+
     // Access QUnit internals to get all suites and tests, working around
     // missing event data.
 
@@ -128,6 +142,9 @@ export default class QUnitAdapter extends EventEmitter {
 
         var parentSuiteName = suite.name.substring(indexStart, indexEnd)
 
+        // Keep only the name of the suite itself.
+        suite.name = suite.name.substring(indexEnd + 3)
+
         for (let parentSuite of suites) {
           if (parentSuite.name === parentSuiteName) {
             parentSuite.childSuites.push(suite)
@@ -139,6 +156,8 @@ export default class QUnitAdapter extends EventEmitter {
     }
 
     globalSuite.childSuites = topLevelSuites
+
+    this.emitGlobal = false
 
     return globalSuite
   }
--- a/lib/adapters/QUnitAdapter.js
+++ b/lib/adapters/QUnitAdapter.js
@@ -40,6 +40,7 @@ export default class QUnitAdapter extends EventEmitter {
 
     if (qunitTest.errors !== undefined) {
       var status
+      var runtime
 
       if (qunitTest.failed > 0) {
         status = 'failed'
@@ -49,7 +50,11 @@ export default class QUnitAdapter extends EventEmitter {
         status = 'passed'
       }
 
-      return new Test(qunitTest.name, suiteName, status, qunitTest.runtime,
+      if (status !== 'skipped') {
+        runtime = qunitTest.runtime
+      }
+
+      return new Test(qunitTest.name, suiteName, status, runtime,
           qunitTest.errors)
     }
 
--- a/lib/adapters/QUnitAdapter.js
+++ b/lib/adapters/QUnitAdapter.js
@@ -12,22 +12,52 @@ export default class QUnitAdapter extends EventEmitter {
     QUnit.testDone(this.onTestDone.bind(this))
     QUnit.moduleDone(this.onModuleDone.bind(this))
     QUnit.done(this.onDone.bind(this))
-
-    this.idDataMap = {}
   }
 
   convertModule (qunitModule) {
     var suiteName = qunitModule.name
     var tests = []
+
     for (let qunitTest of qunitModule.tests) {
-      let test = new Test(qunitTest.name, suiteName)
-      tests.push(test)
-      this.idDataMap[qunitTest.testId] = test
+      tests.push(this.convertTest(qunitTest, suiteName))
     }
 
-    var suite = new Suite(suiteName, [], tests)
-    this.idDataMap[qunitModule.name] = suite
-    return suite
+    return new Suite(suiteName, [], tests)
+  }
+
+  /**
+   * @param suiteNameParam is used only when the test does not contain the
+   * parent module name, if it is available in the test, then it is undefined.
+   *
+   * The parent module name is not available when converting the suites on
+   * the "runtStart" event.
+   */
+  convertTest (qunitTest, suiteNameParam) {
+    var suiteName = suiteNameParam
+
+    // Assign the name of the module, only if it is not the global one, then
+    // leave the suiteName to be undefined.
+    if (!suiteName && qunitTest.module !== '') {
+      suiteName = qunitTest.module
+    }
+
+    if (qunitTest.errors !== undefined) {
+      var status
+
+      if (qunitTest.failed > 0) {
+        status = 'failed'
+      } else if (qunitTest.skipped) {
+        status = 'skipped'
+      } else {
+        status = 'passed'
+      }
+
+      return new Test(qunitTest.name, suiteName, status, qunitTest.runtime,
+          qunitTest.errors)
+
+    }
+
+    return new Test(qunitTest.name, suiteName)
   }
 
   onBegin () {
@@ -38,19 +68,19 @@ export default class QUnitAdapter extends EventEmitter {
       suites.push(this.convertModule(suite))
     }
 
-    this.idDataMap.globalSuite = new Suite(undefined, suites, [])
-    this.emit('runStart', this.idDataMap.globalSuite)
+    this.globalSuite = new Suite(undefined, suites, [])
+    this.emit('runStart', this.globalSuite)
   }
 
   onModuleStart (details) {
     if (details.name !== '') {
-      this.emit('suiteStart', this.idDataMap[details.name])
+      this.emit('suiteStart', this.convertModule(details))
     }
   }
 
   onTestStart (details) {
     this.errors = []
-    this.emit('testStart', this.idDataMap[details.testId])
+    this.emit('testStart', this.convertTest(details))
   }
 
   onLog (details) {
@@ -60,29 +90,23 @@ export default class QUnitAdapter extends EventEmitter {
   }
 
   onTestDone (details) {
-    var test = this.idDataMap[details.testId]
+    var test
 
-    if (details.failed !== 0) {
-      test.status = 'failed'
-    } else if (details.skipped === true) {
-      test.status = 'skipped'
-    } else {
-      test.status = 'passed'
-    }
-
-    test.runtime = details.runtime
-    test.errors = this.errors
+    details.errors = this.errors
+    test = this.convertTest(details)
 
     this.emit('testEnd', test)
   }
 
   onModuleDone (details) {
     if (details.name !== '') {
-      this.emit('suiteEnd', this.idDataMap[details.name])
+      this.emit('suiteEnd', this.convertModule(details))
+    } else {
+      this.globalEndSuite = this.convertModule(details)
     }
   }
 
   onDone () {
-    this.emit('runEnd', this.idDataMap.globalSuite)
+    this.emit('runEnd', this.globalEndSuite)
   }
 }
--- a/lib/adapters/QUnitAdapter.js
+++ b/lib/adapters/QUnitAdapter.js
@@ -43,7 +43,9 @@ export default class QUnitAdapter extends EventEmitter {
   }
 
   onModuleStart (details) {
-    this.emit('suiteStart', this.idDataMap[details.name])
+    if (details.name !== '') {
+      this.emit('suiteStart', this.idDataMap[details.name])
+    }
   }
 
   onTestStart (details) {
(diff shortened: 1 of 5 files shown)


lib/adapters/QUnitAdapter.js
```diff
@@ -58,27 +88,23 @@ export default class QUnitAdapter extends EventEmitter {
...
-    } else if (details.skipped === true) {
-      test.status = 'skipped'
-    } else {
-      test.status = 'passed'
-    }
+    var test
 
-    test.runtime = details.runtime
-    test.errors = this.errors
+    details.errors = this.errors
```
flore77: I have done this to equal the MochaAdapter, here we can check in the convertTest function for `this.errors`

flore77:
@jzaefferer I know the tests are still failing, but I want to have some feedback for this part.

There are still 2 big issues to solve: `suiteEnd objects` and `nested suites`.

Here is does not function the trick to add to the test object the `errors` property, because QUnit is emitting other objects on each event. So on the `suiteEnd` event, the emitted objects will contain a list of `tests`, but this have only 2 props: `name` and `testId`. We need also `runtime`, `error` and `status`.

I am thinking to index the aforementioned props into an array, something:

```js
var testProps = {}

var props = {
  status: test.status,
  errors: test.errors,
  runtime: test.runtime
}

testProps[test.testId] = props. 
```
So in the `convertTest` function, if we don't have the all the props we need, we can check for them in the `testProps` object.

lib/adapters/QUnitAdapter.js
```diff
@@ -12,22 +12,50 @@ export default class QUnitAdapter extends EventEmitter {
...
+    return new Suite(suiteName, [], tests)
+  }
+
+  /**
+   * @param qunitTest - when passed in from the convertModule function it does
+   *  contain only its "name" and the "testId".
+   * @param suiteNameParam - is used only when calling convertTest from
+   *  the convertModule function, in rest it is undefined.
+   */
+  convertTest (qunitTest, suiteNameParam) {
```
jzaefferer: `suiteNameParam` is a weird parameter, since its used only once. I suggest renaming this to `suiteName`, then overwriting this as necessary.

lib/adapters/QUnitAdapter.js
jzaefferer: Is this just for consistency, or is it actually needed?

lib/adapters/QUnitAdapter.js
```diff
@@ -12,22 +12,50 @@ export default class QUnitAdapter extends EventEmitter {
...
+    var suiteName = suiteNameParam
+
+    // Assign the name of the module, only if it is not the global one, then
+    // leave the suiteName to be undefined.
```
jzaefferer: This is a weird comment, let's try to rephrase this.

lib/adapters/QUnitAdapter.js
jzaefferer: I guess QUnit also doesn't provide aggregated errors, so we need that here.

lib/adapters/QUnitAdapter.js
```diff
@@ -58,27 +88,23 @@ export default class QUnitAdapter extends EventEmitter {
...
-    test.errors = this.errors
+    details.errors = this.errors
+    test = this.convertTest(details)
 
     this.emit('testEnd', test)
   }
 
   onModuleDone (details) {
-    this.emit('suiteEnd', this.idDataMap[details.name])
+    if (details.name !== '') {
```
jzaefferer: This isn't a great check, and its now scattered across multiple places in this class. Let's try to at least abstract it, so that we can call `if (this.globalModule(details)) { [...]`.

test/fixtures/qunit.js
```diff
@@ -1,12 +1,16 @@
 var QUnit = require('qunitjs')
 
-QUnit.test('global test', function () {
+// Add dummy assertions in passing tests, because QUnit is checking for at
+// least one assertion, if there is none, then QUnit throws an error and the
+// test fails.
 
+QUnit.test('global test', function () {
+  QUnit.assert.ok(true)
```
jzaefferer: You should define the `assert` argument and use:

```
QUnit.test('global test', function (assert) {
  assert.ok(true)
})
```

jzaefferer:
> So in the convertTest function, if we don't have the all the props we need, we can check for them in the testProps object.

Let's give that a try.

lib/adapters/QUnitAdapter.js
```diff
@@ -58,27 +111,21 @@ export default class QUnitAdapter extends EventEmitter {
...
-    test.errors = this.errors
-
-    this.emit('testEnd', test)
+    details.testEnd = true
+    this.emit('testEnd', this.convertTest(details))
   }
 
   onModuleDone (details) {
-    this.emit('suiteEnd', this.idDataMap[details.name])
+    details.suiteEnd = true
```
flore77: Notify `convertModule` it must emit a `suite end`.

lib/adapters/QUnitAdapter.js
```diff
@@ -58,27 +111,21 @@ export default class QUnitAdapter extends EventEmitter {
...
-      test.status = 'skipped'
-    } else {
-      test.status = 'passed'
-    }
-
-    test.runtime = details.runtime
-    test.errors = this.errors
-
-    this.emit('testEnd', test)
+    details.testEnd = true
```
flore77: Notify `convertTest` it must emit a `test end`.

lib/adapters/QUnitAdapter.js
```diff
@@ -13,21 +13,72 @@ export default class QUnitAdapter extends EventEmitter {
...
 
   convertModule (qunitModule) {
     var suiteName = qunitModule.name
     var tests = []
+
     for (let qunitTest of qunitModule.tests) {
-      let test = new Test(qunitTest.name, suiteName)
-      tests.push(test)
-      this.idDataMap[qunitTest.testId] = test
+      qunitTest.testEnd = qunitModule.suiteEnd
```
flore77: If a `suite end` must be emitted, notify also `converTest` to emit a `test end`.

flore77:
@jzaefferer please review.

lib/adapters/QUnitAdapter.js
jzaefferer: I don't understand what you're talking about. There are no `emit` calls in `convertTest`

lib/adapters/QUnitAdapter.js
jzaefferer: Same as above. Your comment doesn't reflect the implementation. And I don't think we should be adding even more properties to the data we're consuming. We really should consider those objects immutable and track these properties ourselves, like we now do with `this.testsDetails`.

lib/adapters/QUnitAdapter.js
flore77: Yes, sorry, it should to **create** a `test end`.

jzaefferer: Can you split `convertTest` into 2 methods, one doing the basic conversion, the other doing the `testEnd` work? Then `onTestDone` call both methods.

flore77: Hmm. Something like:

```js
function saveTestDetails(qunitTest) {
  var testDetails = {}
   // attach the details to the above object
   this.testsDetails[qunitTest.testId] = testDetails
}

function convertTest(qunitTest, suiteName) {
  var testDetails = this.testsDetails[qunitTest.testId]
  
  if (testDetails !== undefined) {
     // emit a test end
  }

  // emit a test start
}

function onTestDone(details) {
  saveTestDetails(qunitTest);
  this.emit('testEnd', convertTest(qunitTest));
}
```

Is this ok ?

jzaefferer: Yeah, that looks better.

flore77:
Done.

jzaefferer:
That looks better. `npm test` is failing for me with this output:

```
/Users/jzaefferer/dev/jquery/js-reporters/dist/js-reporters.js:1175
          if (actual.name !== expected.name) {
                    ^

TypeError: Cannot read property 'name' of undefined
```
Haven't tried to debug that, yet, are you aware of this issue?

flore77:
Yes, I have seen it, but I still wanted to push, because the integration tests are passing, at least some of them that should pass :laughing: 

> Haven't tried to debug that, yet

I will debug it!
 
Maybe it gets `undefined`, because now there are less events, we do not emit the `globalSuite` twice.

flore77:
I found the problem and it is a bug :disappointed: The `globalModule` is not emitted if it does `not contain` any `tests`, so on the `runEnd` event `this.globalSuiteEnd` will be undefined. In our test fixture we have a global test so everything is functioning well.

flore77:
I would leave this after we found a fix for the nested suites, because also now the suite on the runStart event is not created quite well. After fixing the nesting suites, it would be more clearer how it must be built.

jzaefferer:
I didn't understand your explanations. Let's make `npm test` pass, even if that means skipping a few tests temporarily.

flore77:
The [onBegin](https://github.com/js-reporters/js-reporters/blob/39c02f654ec4ff52de417e5255863885677f7980/lib/adapters/QUnitAdapter.js#L86-L90) function does not emit a suite that respects the standard, because:

- it does not treat nested suites, i.e the `outter suite` has `no childSuites` (it should have the inner suite), `all suites` are putted as `childSuites` of the `globalSuite`
- QUnit global module is wrapped in an another module instead of being the one that wraps all other modules

The [moduleDone](https://github.com/js-reporters/js-reporters/blob/39c02f654ec4ff52de417e5255863885677f7980/lib/adapters/QUnitAdapter.js#L120) functions saves in `this.globalEndSuite` a converted suite that does not contain the other suites, something of this form:

```js
Suite {
  name: '',
  childSuites: [],
  tests: 
   [ Test {
       testName: 'global test',
       suiteName: undefined,
       status: 'passed',
       runtime: 1,
       errors: [] } ] }
```

I was thinking to firstly fix the nested modules, and then to build the `suites` emitted on `runStart` and `runEnd`, because then we will have the big picture, while now we don't now how we will deal with the nested suites.

The `undefined` comes from, because the old test fixture does not have a global test and QUnit does not emit the globalModule anymore on the `suiteStart` and `suiteEnd` events. And so our adapter will never enter on the [else](https://github.com/js-reporters/js-reporters/blob/39c02f654ec4ff52de417e5255863885677f7980/lib/adapters/QUnitAdapter.js#L119), so `this.globalEndSuite` will stay undefined. 

I hope I have explained more clearly :smiley:  

How should I proceed ? Should I know to take care of building the `suites` properly for the `runStart` and `runEnd` events ? Or should I deal with the nested modules ?

jzaefferer:
> The undefined comes from, because the old test fixture does not have a global test and QUnit does not emit the globalModule anymore on the suiteStart and suiteEnd events. And so our adapter will never enter on the else, so this.globalEndSuite will stay undefined.

Is there a simple fix or workaround for that, so that we can make these tests pass while working on other issues?

> How should I proceed ? Should I know to take care of building the suites properly for the runStart and runEnd events ? Or should I deal with the nested modules ?

Either seems fine to me, I can't tell which one is better to start with. Pick one, if you get stuck, try the other?

flore77:
@jzaefferer I have fixed the global suites emitted on the runStart/runEnd events, I am looking forward for your feedback :smiley: 

The 2 tests that are testing this events are still failing, but because of the `name` of the `inner suite`. 

QUnit emits also for suites a concatenated name, the inner suite has the name: `outter suite > inner suite`and in our reference data it has the name `inner suite`. How should this be ?

The inner tests has also a concatenated name, but we must only remove the `>` character. I am waiting to know how we deal with the suite name, then I will solve them both.

jzaefferer:
I'll have to review the `createGlobalSuite` more thorougly later, at a glance it looks good. Nice job!

> QUnit emits also for suites a concatenated name, the inner suite has the name: outter suite > inner suiteand in our reference data it has the name inner suite. How should this be ?

> The inner tests has also a concatenated name, but we must only remove the > character. I am waiting to know how we deal with the suite name, then I will solve them both.

This feels like something we should bring back to QUnit. For now let's just cut of the last `>` and anything before that. That should be simple enough to adjust later.

flore77:
Now remains only to solve the execution of nested modules.

I have seen that also Jasmine executes the nested suites this way, from the inner most to the outer most suite.

How should we do this in our standard ?

Also from what I have seen, QUnit does not emit the start of the `outter suite` and then the start of the `inner suite`, it executes them indepedently, i.e `inner suite starts`, `inner test executes`, `inner suite ends`,  `outer suite starts`, `outer test executes`, `outer suite ends`, so any option we choose, we will have some code to write for the QUnitAdapter.

jzaefferer:
Are the Jasmine and QUnit behaviours at least consistent? Or do they also have differences?

> so any option we choose, we will have some code to write for the QUnitAdapter.

I have no preference, what do you think?

flore77:
> Are the Jasmine and QUnit behaviours at least consistent? Or do they also have differences?

Yes, there are some small differences, but I didn't investigate too deep for Jasmine.

I think the most natural is to run the tests from the outer most suite to the inner most suite, like Mocha does.  Also the **QUnit reporter displays tests in the same order as Mocha**, no matter in which order the tests are executed and emitted.

So, I think definitely to emit the tests from the outer most to the inner most.

I will try today to achieve this for QUnit, I am thinking to buffer all data in our adapter until QUnits calls the done callback and after that to begin emit data from our adapter. This way we can store and order easily the data in a form we like. Sounds ok ?

jzaefferer:
Yes, that sounds good. Could you also file an issue against QUnit, telling them about the difference with Mocha, and the inconsistency in the UI? Maybe they're interested in addressing that before the 2.0.0 release.

jzaefferer:
Before you work on the suites handling, could we skip those 8 failing tests and land this PR? Then work on the suites handling in a separate PR.

flore77:
>  Could you also file an issue against QUnit, telling them about the ...

Hmm, I don't know, QUnit has  2 options that randomizes tests execution:
* the `reorder` option that on a rerun, runs firstly the failed tests, it is activated by default
* the `seed` option that randomizes tests execution, it is disabled by default

The `reorder` option is activated by default and in it description it says:
> this can speed up testing a lot.
It can also lead to random errors, in case your testsuite has non-atomic tests, where the order is important. You should fix those issues, instead of disabling reordering!`

So, I think it is an important feature of QUnit and I think in their reporter they are doing this reordering, because it would be confusing on multiple runs to display tests in a different order.

Of course the inconsistency is that on a normal running they are running the tests from the inner most to the outer most, but displaying them the other way, but I think the reporter cannot be change, because it must handle the other cases of `reorder` and `seed`.  So the change that can be done, would be change the normal running, but I don't know if it would be a gain. 

What do you think? Should I open an issue there to see what they say about it ?

jzaefferer:
Thanks for the details, I hadn't considered those. So no, let's not do that now.

As for this PR, let's skip the remaining failing tests and merge.

flore77:
Yeah, we have a problem, the `rollup-plugin-babel` is causing this: 

[Cannot find module '__babelHelpers__' from '/home/flore77/js-reporters/node_modules/events'](https://github.com/rollup/rollup-plugin-babel/blob/master/CHANGELOG.md), I don't know yet how to solve this, I have forced to stay for now at the 2.4.0.

flore77:
Maybe we should change the assertion of the runtime range, to check only if runtime is a number, like:

`expect(typeof runtime).to.be.equal('number')`.

jzaefferer:
> Maybe we should change the assertion of the runtime range

Let's see if the updated assertion works out, if not, we could still make that change.

PR #60, opened 2016-06-07
Fix QUnit nested suites

### LINKED ISSUE #57: Fix QUnit Adapter

flore77:
Fixes #57, [6] nested modules.

### DIFF
--- a/lib/adapters/QUnitAdapter.js
+++ b/lib/adapters/QUnitAdapter.js
@@ -12,10 +12,6 @@ export default class QUnitAdapter extends EventEmitter {
     QUnit.done(this.onDone.bind(this))
 
     this.testsDetails = {}
-    this.type = {
-      START: 'S',
-      END: 'E'
-    }
   }
 
   convertModule (qunitModule) {
@@ -129,48 +125,43 @@ export default class QUnitAdapter extends EventEmitter {
     return globalSuite
   }
 
-  createSuite (suite, type) {
-    var childSuites = []
-    var tests = []
-
-    for (let childSuite of suite.childSuites) {
-      childSuites.push(this.createSuite(childSuite, type))
-    }
-
-    for (let test of suite.tests) {
-      tests.push(this.createTest(test, type))
-    }
-
-    return new Suite(suite.name, childSuites, tests)
+  createSuiteStart (suite) {
+    return new Suite(
+      suite.name,
+      suite.childSuites.map(this.createSuiteStart.bind(this)),
+      suite.tests.map(this.createTestStart.bind(this))
+    )
   }
 
-  createTest (test, type) {
-    switch (type) {
-      case this.type.START: {
-        return new Test(test.testName, test.suiteName)
-      }
-      case this.type.END: {
-        var testDetails = this.testsDetails[test.testId]
+  createSuiteEnd (suite) {
+    return new Suite(
+      suite.name,
+      suite.childSuites.map(this.createSuiteEnd.bind(this)),
+      suite.tests.map(this.createTestEnd.bind(this))
+    )
+  }
 
-        return new Test(test.testName, test.suiteName, testDetails.status,
-          testDetails.runtime, testDetails.errors)
-      }
-      default: {
-        throw new Error('Unknown test type')
-      }
-    }
+  createTestStart (test) {
+    return new Test(test.testName, test.suiteName)
+  }
+
+  createTestEnd (test) {
+    var testDetails = this.testsDetails[test.testId]
+
+    return new Test(test.testName, test.suiteName, testDetails.status,
+      testDetails.runtime, testDetails.errors)
   }
 
   emitData (suite) {
     for (let test of suite.tests) {
-      this.emit('testStart', this.createTest(test, this.type.START))
-      this.emit('testEnd', this.createTest(test, this.type.END))
+      this.emit('testStart', this.createTestStart(test))
+      this.emit('testEnd', this.createTestEnd(test))
     }
 
     for (let suite of suite.childSuites) {
-      this.emit('suiteStart', this.createSuite(suite, this.type.START))
+      this.emit('suiteStart', this.createSuiteStart(suite))
       this.emitData(suite)
-      this.emit('suiteEnd', this.createSuite(suite, this.type.END))
+      this.emit('suiteEnd', this.createSuiteEnd(suite))
     }
   }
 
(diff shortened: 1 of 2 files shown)


flore77:
@jzaefferer please take a look :smiley: 

We must do something with the tests runtime range test, `f2b23c8` failed for that reason.

jzaefferer:
Sucks that we can only emit at the end of the run, but that's certainly a good reason to get the reporter implemented natively in QUnit itself asap.

lib/adapters/QUnitAdapter.js
```diff
@@ -6,67 +6,37 @@ export default class QUnitAdapter extends EventEmitter {
...
-    QUnit.moduleStart(this.onModuleStart.bind(this))
     QUnit.testStart(this.onTestStart.bind(this))
     QUnit.log(this.onLog.bind(this))
     QUnit.testDone(this.onTestDone.bind(this))
-    QUnit.moduleDone(this.onModuleDone.bind(this))
     QUnit.done(this.onDone.bind(this))
 
     this.testsDetails = {}
-    this.emitGlobal = false
+    this.type = {
```
jzaefferer: I don't like this object, same issue as the event name abstraction we discussed a while ago. Passing `'start'` and `'end'`as strings will do the trick.

jzaefferer: Let's try to get rid of the `type` argument entirely, and use separate `create` methods instead. `createTestStart` and `createTestEnd` should be obvious enough to implement. As for the other two, something like this, simplified using `map` instead of `let of / push`:

```js
createSuiteStart () {
  return new Suite(suite.name,
    suite.childSuites.map(this.createSuiteStart),
    suite.tests.map(this.createTestStart))
}

createSuiteEnd () {
  return new Suite(suite.name,
    suite.childSuites.map(this.createSuiteEnd),
    suite.tests.map(this.createTestEnd))
}
```
May need to explicitly bind both functions inside the constructor, to be able to pass them to `map` and still get the same `this` reference.

lib/adapters/QUnitAdapter.js
```diff
@@ -82,40 +52,39 @@ export default class QUnitAdapter extends EventEmitter {
...
+      modules = this.QUnit.config.modules
     }
 
     // Build a list with all suites.
-    for (let module of this.QUnit.config.modules) {
-      if (!this.isGlobalModule(module)) {
-        suites.push(this.convertModule(module))
-      }
+    for (let module of modules) {
+      suites.push(this.convertModule(module))
```
jzaefferer: Should also use `map` here.

flore77:
Done.

lib/adapters/QUnitAdapter.js
```diff
@@ -82,41 +48,38 @@ export default class QUnitAdapter extends EventEmitter {
...
+      modules = this.QUnit.config.modules
     }
 
     // Build a list with all suites.
-    for (let module of this.QUnit.config.modules) {
-      if (!this.isGlobalModule(module)) {
-        suites.push(this.convertModule(module))
-      }
-    }
+    suites = modules.map(this.convertModule)
```
jzaefferer: Now that we don't need the initializer, let's remove the declaration above, and use `let suites =...` instead. Check if that can be applied for other vars as well.

lib/adapters/QUnitAdapter.js
```diff
@@ -157,24 +120,55 @@ export default class QUnitAdapter extends EventEmitter {
...
+
+  createTestStart (test) {
+    return new Test(test.testName, test.suiteName)
   }
 
-  onModuleStart (details) {
-    if (!this.isGlobalModule(details)) {
-      this.emit('suiteStart', this.convertModule(details))
+  createTestEnd (test) {
+    var testDetails = this.testsDetails[test.testId]
```
jzaefferer: Give this a try:
```
const { status, runtime, errors } = this.testsDetails[test.testId]
return new Test(test.testName, test.suiteName, status, runtime, errors)
```

flore77:
Can this be landed?

jzaefferer:
👍

PR #66, opened 2016-06-19
Enhance mocha adapter

flore77:
@jzaefferer please review

### DIFF
--- a/lib/adapters/MochaAdapter.js
+++ b/lib/adapters/MochaAdapter.js
@@ -1,13 +1,15 @@
 import EventEmitter from 'events'
-import { Test, Suite } from '../Data.js'
+import {Test, Suite} from '../Data.js'
 
 export default class MochaAdapter extends EventEmitter {
   constructor (mocha) {
     super()
 
     this.mocha = mocha
+
     mocha.reporter((runner) => {
       this.runner = runner
+
       runner.on('start', this.onStart.bind(this))
       runner.on('suite', this.onSuite.bind(this))
       runner.on('test', this.onTest.bind(this))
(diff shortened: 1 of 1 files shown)


lib/adapters/MochaAdapter.js
```diff
@@ -1,16 +1,19 @@
...
 
     this.mocha = mocha
+
     mocha.reporter((runner) => {
       this.runner = runner
+
       runner.on('start', this.onStart.bind(this))
       runner.on('suite', this.onSuite.bind(this))
       runner.on('test', this.onTest.bind(this))
+      runner.on('pending', this.onPending.bind(this))
```
jzaefferer: That's a category of change I expect to be covered by tests. Since you didn't modify any tests, there must be missing coverage, right?

jzaefferer:
Besides the missing test coverage, this looks good. Brings the adapter in line with improvements we applied in the others. 👍

lib/adapters/MochaAdapter.js
flore77: I think it is covered, because there are 2 skipped tests in Mocha's test fixture. Through this change I only moved [some code](https://github.com/js-reporters/js-reporters/pull/66/commits/7c56c0aaa20dc84b2a8f45232b63c737fcb72526) from `onTestEnd` function into the new `onPending` function.

flore77: @jzaefferer are you ok with this?

jzaefferer: Yep, sounds good. This looked like more than a refactoring at first.

jzaefferer:
Let's land it.

PR #69, opened 2016-06-23
TapReporter: update to be TAP13 compliant.

### LINKED ISSUE #67: Update TAP reporter
Update TAP reporter to be [TAP13](https://testanything.org/tap-version-13-specification.html) compliant.

flore77:
Fixes #67.

### DIFF
--- a/test/unit/tap-reporter.js
+++ b/test/unit/tap-reporter.js
@@ -17,47 +17,43 @@ describe('Tap reporter', function () {
     JsReporters.TapReporter.init(emitter)
   })
 
-  it('should output the TAP header', function () {
-    var spy = sinon.stub(console, 'log')
+  it('should output the TAP header', sinon.test(function () {
+    var spy = this.stub(console, 'log')
 
     emitter.emit('runStart', {})
-    console.log.restore()
 
     expect(spy).to.have.been.calledOnce
-  })
+  }))
 
-  it('should output ok for a passing test', function () {
-    var spy = sinon.stub(console, 'log')
+  it('should output ok for a passing test', sinon.test(function () {
+    var spy = this.stub(console, 'log')
     var expected = 'ok 1 ' + data.passingTest.testName
 
     emitter.emit('testEnd', data.passingTest)
-    console.log.restore()
 
     expect(spy).to.have.been.calledWith(expected)
-  })
+  }))
 
-  it('should output ok for a skipped test', function () {
-    var spy = sinon.stub(console, 'log')
+  it('should output ok for a skipped test', sinon.test(function () {
+    var spy = this.stub(console, 'log')
     var expected = 'ok 2 ' + data.skippedTest.testName + ' # SKIP'
 
     emitter.emit('testEnd', data.skippedTest)
-    console.log.restore()
 
     expect(spy).to.have.been.calledWith(expected)
-  })
+  }))
 
-  it('should output not ok for a failing test', function () {
-    var spy = sinon.stub(console, 'log')
+  it('should output not ok for a failing test', sinon.test(function () {
+    var spy = this.stub(console, 'log')
     var expected = 'not ok 3 ' + data.failingTest.testName
 
     emitter.emit('testEnd', data.failingTest)
-    console.log.restore()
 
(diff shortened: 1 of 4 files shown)


jzaefferer:
👍 

Would be interesting to see some integration tests, e.g. with node-tap's CLI reporter.

flore77:
> Would be interesting to see some integration tests, e.g. with node-tap's CLI reporter.

What do you mean by `node-tap's CLI reporter` ?

I was thinking maybe some unit tests.

In the reporter constructor to have something like:
```js
constructor(runner, output) {
  ...
  this.write = output || console.log;
}

onRunStart() {
  this.write('TAP version 13');
}

...
```

This way we can pass in:
* an EventEmitter for the runner 
* for output a `spy`

This way we can emit events and we can check the spy is called with the correct parameters, something like:
```js
var ev

before(function() {
  ev = new EventEmitter();
  var reporter = new TapReporter(ev, spy);
});

it('should have the TAP13 header', function() {
  ev.emit('runStart', {});
  
  expect(spy).to.have.been.calledWith('TAP version 13');
}); 
```

Of course we can do, instead (or both), some integration testing:
* instantiate one of our adapters
* run it against its fixture
* use as reporter the TapReporter
* collect the reporter output
* compare the output with the expected one

jzaefferer:
node-tap: https://github.com/tapjs/node-tap - has a nice CLI reporter, based on tap output.

I'll respond to the rest later

jzaefferer:
Instead of modifying the constructor, can we stub `console.log` instead?

Testing that a static string is emitted on `runStart` doesn't seem useful, that would only ever change when updating the TAP version, and you'd also have to update the test.

I'd focus on logic inside the tap reporter has has some complexity, here obviously the `onTestEnd` method. You can create an instance, stub `console.log`, then call `onTestEnd` directly, and verify the log calls.

flore77:
I added some tests please check them out :smiley:

test/unit/tap-reporter.js
```diff
@@ -0,0 +1,87 @@
...
+    var spy = sinon.stub(console, 'log')
+
+    emitter.emit('runStart', {})
+    console.log.restore()
+
+    expect(spy).to.have.been.calledOnce
+  })
+
+  it('should output ok for a passing test', function () {
+    var spy = sinon.stub(console, 'log')
```
flore77: I must do this stubbing in each test and then the restoring below, because otherwise Mocha's reporter output will be also captured and will not be outputted to the console.

test/unit/tap-reporter.js
```diff
@@ -0,0 +1,87 @@
...
+var expect = chai.expect
+
+chai.use(require('sinon-chai'))
+
+describe('Tap reporter', function () {
+  var emitter
+
+  before(function () {
+    emitter = new EventEmitter()
+    var reporter = new JsReporters.TapReporter(emitter)
```
flore77: I don't know how to go over this linting issue. Help wanted :smile:

jzaefferer: That indicates that we shouldn't be using a constructor here. Calling `JsReporters.TapReporter` without `new` seems fine.

test/unit/tap-reporter.js
jzaefferer: Usually it helps to call `sandbox = sinon.sandbox.create()` in a `beforeEach`, then create whatever stub you need per test using `sandbox.stub`, and call `sandbox.restore()` once in an `afterEach`.

Or move the console.log stub call into a beforeEach, and group tests that need that into a nested suite.

flore77: It will not work with `beforeEach` and `afterEach`, because we will miss Mocha's reporter output for the test that has been exectued.

I think I found a way by wrapping the test function in `sinon.test`, by wrapping it up a sandbox is created on the function call and it is automatically restored on its end, this way we will not miss the reporter output. I will update the tests in 1 min.

flore77:
Done.

jzaefferer:
This looks good.

flore77:
Land? :smiley:

jzaefferer:
Yes, that's what I meant :-)

PR #70, opened 2016-06-26
Test the dynamic props of the suite objects

### LINKED ISSUE #68: Test suite's status and runtime props
Currently the testing does not verify the `status` and `runtime` props of `suites`.

This props are implemented through ES6 getters, so they have dynamically computed values.

I see 2 ways to test the props:

1 - To make the properties constant

2 - To create 2 new tests for this props

```js
it('test the suites statuses', function() {
  suites.forEach(function(suite, index) {
    expect(suite.status).to.be.equal(expectedSuitesStatus[index]);
  });
});

it('test the suites runtime to be a Number', function() {
  ...
});
```

I am open to other suggestions.

flore77:
Fixes #68.


flore77:
I think this is the easiest way to test the dynamic props.

They already uncovered a bug regarding the `runtime` when the suite is containing skipped tests.

flore77:
Although I am not sure, how the `runtime` should be for a suite with `skipped status`. Currently it will be 0, should it be undefined?

jzaefferer:
To be consistent, `undefined` seems the right choice. Though I wonder if we should revert that whole thing and go with QUnit's behaviour, with `0` for skipped tests runtime... Might be a lot easier overall.

jzaefferer:
Otherwise this is looking good.

flore77:
> Though I wonder if we should revert that whole thing and go with QUnit's behaviour, with 0

Yes, we could, maybe we will address this in another issue.

Good to land ?

jzaefferer:
Yeah, good to land

PR #74, opened 2016-07-07
Run also Mocha's default reporter

flore77:


flore77:
@jzaefferer please review. The tests are not passing because of [this unused var](https://github.com/js-reporters/js-reporters/pull/74/commits/65d42595b0acace1b722e4b57754eb6ac5301cb9#diff-f024e622c47dd0392896f164ba0caa35R15). I don't know if it can be done without `new`, I have tried to do it with apply and call, but I didn't succeed, I think it must be done with new, because Mocha's default reporters are inheriting the Base reporter. Can a class/object be instantiated without new?

jzaefferer:
You can suppress that error. Do it inline so that it affects only the one line: http://eslint.org/docs/user-guide/configuring#disabling-rules-with-inline-comments

lib/adapters/MochaAdapter.js
```diff
@@ -6,10 +6,14 @@ export default class MochaAdapter extends EventEmitter {
     super()
 
     this.mocha = mocha
+    this.origReporter = mocha._reporter
 
     mocha.reporter((runner) => {
       this.runner = runner
 
+      let constructor = this.origReporter.bind(this.mocha, this.runner)
```
jzaefferer: Can you we `new this.origReporter.bind(this.mocha, this.runner)()` here? Along with suppressing whatever eslint rule will fail it.

flore77:
Done. Mocha's output is still visible in the old testing style, but I cannot suppress it like I did in `adapters-run.js`, because it will also suppress the errors logged by the TestReporter.

jzaefferer:
Does it make sense to keep the old tests? If not, we might as well remove them now.

Either way, this looks good to land.

flore77:
> Does it make sense to keep the old tests? If not, we might as well remove them now.

No, I don't think it's make sens to keep them, but we will remove them on another branch, maybe when we will run our tests also in the browser.

PR #75, opened 2016-07-09
First release

flore77:


flore77:
@jzaefferer should I add something else to `package.json` or `.npmignore` files?

flore77:
Also should we release, `0.0.0` or `0.0.1` version, because we are still developing. Or `1.0.0` ?

.npmignore
```diff
@@ -0,0 +1,2 @@
+.travis.yml
+.npmignore
```
jzaefferer: Is this necessary? Seems like npm should ignore this file by default.

jzaefferer:
To test the publishing process, you can change the name to the @username/package format, to publish to your npm user namespace. Then unpublish that afterwards and revert the name change.

I suggest we start with 1.0.0. The browserstack-runner integration hasn't surfaced any major problems, so might as well show some confidence.

PS: I'd squash these changes into a single commit.

.npmignore
flore77: Hmm, no, it is not ignoring it by default. I tested and without this line it will get also the `.npmignore` file.

flore77:
> To test the publishing process, you can change the name to the @username/package format, to publish to your npm user namespace.

I tried and I received this:

> You need a paid account to perform this action. For more info, visit: https://www.npmjs.com/private-modules : @flore77/js-reporters

I tested like described here, in the [npm docs](https://docs.npmjs.com/misc/developers#before-publishing-make-sure-your-package-installs-and-works).

Good to land? After landing, should I publish the package on npm?

jzaefferer:
Yeah, let's land and publish. Can you add me as another owner afterwards? My npm username is jzaefferer

flore77:
Sure.

PR #76, opened 2016-07-12
For...of polyfill

flore77:
Fixes the `for...of` compatibility problem on IE from https://github.com/flore77/browserstack-runner/pull/2.

### DIFF
--- a/lib/Data.js
+++ b/lib/Data.js
@@ -48,8 +48,11 @@ export class Suite {
     var passed = 0
     var failed = 0
     var skipped = 0
+    var tests = this.getAllTests()
 
-    for (let test of this.getAllTests()) {
+    for (let i = 0; i < tests.length; i++) {
+      let test = tests[i]
+
       // If a suite contains a test whose status is still undefined,
       // there is no final status for the suite as well.
       if (test.status === undefined) {
--- a/lib/adapters/JasmineAdapter.js
+++ b/lib/adapters/JasmineAdapter.js
@@ -76,7 +76,7 @@ export default class JasmineAdapter extends EventEmitter {
     var childSuites = []
     var tests = []
 
-    for (let child of jasmineSuite.children) {
+    jasmineSuite.children.forEach((child) => {
       if (child.id.indexOf('suite') === 0) {
         childSuites.push(this.createGlobalSuite(child))
       } else {
@@ -96,7 +96,7 @@ export default class JasmineAdapter extends EventEmitter {
         tests.push(test)
         this.tests[child.id] = test
       }
-    }
+    })
 
     let name = (this.isJasmineGlobalSuite(jasmineSuite)) ? undefined
       : jasmineSuite.description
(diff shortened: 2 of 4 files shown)


flore77:
I tested and now everything is functioning well now, the browserstack tests are passing now also with Mocha's runtime.

I didn't manage it to get it work neither `babel-polyfill`, I get this error:
image

But it does work if I include separately the `dist/polyfill.js` bundle.

leobalter:
for of should not be included by polyfills, this is syntax matters and not a built-in api.

```js
var tests = this.getAllTests()
```

Does it return an array? If so, you can use forEach instead of loops.

If it's an ordinary object and you can use the polyfill for `Object.values`.

flore77:
>  this is syntax matters and not a built-in api.

So and so because `for...of` is compiled by babel with help of `Symbol.iterator`, which is an ES6 feature, so it should be part of the built-in api, but unfortunately is not available in IE and is resulting into this:

image
 
> Does it return an array?
Yes.

>  you can use forEach instead of loops.

Yes, of course, but I don't like that I am not able to use this ES6 feature. And this is not the only place where we are using it and I thought it would be easier to include a polyfill, but I didn't manage to sort it yet.

flore77:
This, https://phabricator.babeljs.io/T7348, is the same error I get when I'm importing `babel-polyfill` in js-reporters.

In `index.js` file, before other imports:
```js
import 'babel-polyfill'
//other imports
```

If I import `babel-polyfill` and `js-reporters` in the browser like 2 separates scripts it works.

flore77:
@jzaefferer I didn't manage to get it work. What should we do? Should we replace all `for...of` ?

jzaefferer:
Yes, let's replace them.

flore77:
Or `babel` or `rollup` is breaking `babel-polyfill`,  I copied the babel-polyfill at the beginning of the js-reporters bundle without including it in the whole bundling process and everything was working. 

So another solution would be this, to change the `rollup.config.js` file to copy babel-polyfill at the beginning of the bundle, or at least the polyfill for `Symbol`.

I don't know which best. What would you prefer ?

jzaefferer:
Any reason against using `forEach` as @leobalter suggested?

flore77:
No, I will update. Only in the `Data.js` file i'll not use it, because there is a return statement and it is not working with forEach, and would have required other refactoring. I think that's I continued with the classical for, sorry, I will update.

jzaefferer:
Okay.

flore77:
Done. After this is merged, we should make another release, so this will arrive also into browserstack. Should we keep a ChangeLog ?

jzaefferer:
Yes, let's make a patch release. You could use GitHub's Releases feature to add a note about the changes.

flore77:
Ok, I haven't used it before, but I don't think it's so hard 😃 Then we can make the pr against browerstack.

PR #77, opened 2016-07-19
Full name prop

### LINKED ISSUE #65: Nested suite name
Currently only tests have as `suiteName` a concatenated name of the nested suites.

Example:

```js
describe('outer suite', function() {
  describe('inner suite', function() {
    it('test', function() {});
  });
});
```

emitted outer suite object:
```js
{
  name: outer suite,
  ...
}
```

emitted inner suite object:
```js
{
  name: inner suite
  ...
}
```

emitted test object:
```js
{
  testName: test,
  suiteName: outer suite inner suite
  ...
}
```

flore77:
Fixes #65 .


jzaefferer:
> I will remove the old testing now, because there is no reason to refactor it.

Ok

> Bow does it look with the canonicalName prop as array?

Looks fine to me. As mentioned on the issue, since you didn't like `canonicalName` much, we can go with `fullName`.

> Also for the globalSuite should it be [undefined] or just [] ?

Empty array seems better in this case.

lib/adapters/MochaAdapter.js
```diff
@@ -29,16 +29,23 @@ export default class MochaAdapter extends EventEmitter {
   convertSuite (mochaSuite) {
     return new Suite(
       mochaSuite.title,
+      this.buildSuiteCanonicalName(mochaSuite),
```
jzaefferer: Now that the property is `fullName`, should update methods still using "canonical" as well.

package.json
```diff
@@ -12,7 +12,7 @@
     "build": "rollup -c",
     "prepublish": "npm run build",
     "pretest": "npm run build",
-    "test": "standard && node test/test.js && npm run test-integration && npm run test-unit",
+    "test": "standard && npm run test-integration && npm run test-unit",
```
jzaefferer: While we're here, lets run unit tests before integration tests.

flore77: Ok!

jzaefferer:
👍

Issue #79, opened 2016-07-20
Normalizing assertions


flore77:
I am thinking something like: 

```js
var assertion = {
  name: String,
  result/successful:  String (failed || passed)/Boolean (true || false),
  actual: Object,
  expected: Object,
  message: String,
  stack: String
};
```

Certainly `actual`, `expected`, `message` and `stack` are mandatory. i am not sure about `name` and `result`.

I think `name`, should be the name of the error, this could be useful, to know if the test failed because of an assertion, or it was the actual code  that generated an error.

`result` could be useful to quickly differentiate between a passed assertion and a failed one.

flore77:
@jzaefferer what do you think about it?

jzaefferer:
What are some example values for `name`?

`result` lacks in clarity. Maybe `passed` or `failed`?

flore77:
> result lacks in clarity

The property can be called or `result` with possible values: `passed` , `failed` or it can be called `succesful` with values: true, false.

For `name` it could be for an error that happend after a failed assertion `AssertionError`. And other errors that can be generated by other code, could have a personalized name, if the thrown error has one, if not it could be assigned a generic name, like `Error`.

And for passed assertion, i don't what value should `name` have. Maybe undefined?

boneskull:
Is this in scope?  This looks like an "assertion library" thing; not a "reporter" thing.

flore77:
Indeed assertions are generated by assertion libraries, but they are reported through the reporting interface and it would be nice if all frameworks would emit the assertions in the same format, so that in reporters would be really easy to display/format/etc. them. An example would be the [browserstack integration](https://github.com/browserstack/browserstack-runner/pull/159/files#diff-885d5623bc9f0fada7aa1bcdcdde5574R24) the part in which the error is formatted.

The error object is quite similar between frameworks, see https://github.com/js-reporters/js-reporters/issues/12#issuecomment-120483356 (scroll down to Error object), so I guess it  wouldn't be such a great effort to standardize it.

boneskull:
OK, I see.  A couple notes around this regarding Mocha:

- It has a notion of a passed test and a failed test of course, but also a "skipped" or "pending" test (this terminology is somewhat confused).  So, a boolean `success` field may then be inappropriate.  I don't know offhand if this "third state" is part of the specification.
- The types of `actual` and `expected` are not defined; these can be primitives or objects.

flore77:
[...]

PR #80, opened 2016-07-20
Test end assertions prop

### LINKED ISSUE #31: testEnd property for all assertions
QUnit reports all assertions, passing or failing: http://api.qunitjs.com/QUnit.log/

There's no need for all frameworks to provide that, but it would make sense to have an array of assertions on the testEnd event. QUnit could make use of that, other frameworks could just leave it null/undefined, reporters can then see if its there and skip it if not.

flore77:
Fixes #31.

### DIFF
--- a/lib/adapters/MochaAdapter.js
+++ b/lib/adapters/MochaAdapter.js
@@ -66,7 +66,7 @@ export default class MochaAdapter extends EventEmitter {
         errors.push(assertion)
       })
 
-      // Test end.
+      // Test end, for the assertions property pass an empty array.
       return new Test(mochaTest.title, suiteName, fullName, status,
           mochaTest.duration, errors, [])
     }
--- a/lib/adapters/MochaAdapter.js
+++ b/lib/adapters/MochaAdapter.js
@@ -52,10 +52,23 @@ export default class MochaAdapter extends EventEmitter {
     // a "test start".
     if (mochaTest.errors !== undefined) {
       var status = (mochaTest.state === undefined) ? 'skipped' : mochaTest.state
+      let errors = []
+
+      mochaTest.errors.forEach(function (error) {
+        let assertion = {
+          passed: false,
+          actual: error.actual,
+          expected: error.expected,
+          message: error.message || error.toString(),
+          stack: error.stack
+        }
+
+        errors.push(assertion)
+      })
 
       // Test end.
       return new Test(mochaTest.title, suiteName, fullName, status,
-          mochaTest.duration, mochaTest.errors)
+          mochaTest.duration, errors, [])
     }
 
     // Test start.
--- a/lib/adapters/QUnitAdapter.js
+++ b/lib/adapters/QUnitAdapter.js
@@ -188,6 +188,7 @@ export default class QUnitAdapter extends EventEmitter {
       actual: details.actual,
       expected: details.expected,
       message: details.message,
+      // For passed assertions the stack must be undefined.
       stack: details.source || undefined
     }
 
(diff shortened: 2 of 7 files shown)


flore77:
I think it would be good to finish first #79. So that we'll know what to expect and to add here proper testing.

lib/adapters/QUnitAdapter.js
```diff
@@ -178,12 +179,23 @@ export default class QUnitAdapter extends EventEmitter {
...
     this.errors = []
+    this.assertions = []
   }
 
   onLog (details) {
-    if (!details.result) {
-      this.errors.push(details)
+    var assertion = {
+      success: details.result,
+      actual: details.actual || null,
```
jzaefferer: This is wrong. Anything falsy is valid here and should not be replaced with `null`. I'd remove the fallback, to keep whatever the property was.

lib/adapters/QUnitAdapter.js
```diff
@@ -178,12 +179,23 @@ export default class QUnitAdapter extends EventEmitter {
...
+      expected: details.expected || null,
```
jzaefferer: Same here

lib/adapters/QUnitAdapter.js
```diff
@@ -178,12 +179,23 @@ export default class QUnitAdapter extends EventEmitter {
...
+      message: details.message,
+      stack: details.source || undefined
```
jzaefferer: This fallback seems unnecessary

flore77: I agree with the ones above, but I think this is necessary, because passed assertion must have the stack undefined, at least this is what we discussed in #79

jzaefferer: Can you verify that `details.source` in QUnit is undefined for passed assertions? If it isn't, we should remove that requirement.

flore77: Ok, I will check 😃

flore77: Yes, it is undefined. More specifically the object emitted on the `log` callback does not have this prop.

flore77: For Jasmine the stack property is an empty string. So, I think `undefined` works just fine.

jzaefferer: Alright. I probably was thinking of QUnit providing stacks for tests, even when passed...

lib/adapters/MochaAdapter.js
```diff
@@ -52,10 +52,25 @@ export default class MochaAdapter extends EventEmitter {
     // a "test start".
     if (mochaTest.errors !== undefined) {
       var status = (mochaTest.state === undefined) ? 'skipped' : mochaTest.state
-
-      // Test end.
+      let errors = []
+
+      // Mocha has indeed only one error per test, but by this is avoided the
```
jzaefferer: There's something wrong with the comment here. "but by this is"? Not sure what you're trying to say.

jzaefferer:
Looks good to me. If we land this as-is, the next step needs to be a spec update. Add the `Error` type next to `Suite` and `Test`.

lib/adapters/MochaAdapter.js
flore77: I wanted to make clear why I used `forEach`, although that Mocha has only one error. I will delete I think.

flore77:
Hmm, `Error` type or better `Assertion` type?

lib/adapters/MochaAdapter.js
jzaefferer: Okay

jzaefferer:
You're right, should be `Assertion`.

PR #82, opened 2016-07-30
Remove getters and setters

flore77:
Fixes https://github.com/browserstack/browserstack-runner/pull/163.


flore77:
I add the regex to replace `babelHelpers.typeof`, because the babel plugin [ES3 member expressions literals transform](http://babeljs.io/docs/plugins/transform-es3-member-expression-literals/) is not transforming also the babel code, it is transforming only our code, so when babel declares:

```js
babelHelpers.typeof = ....
```
This will not be compiled into:
```js
babelHelpers['typeof'] = ....
```

flore77:
By merging #83 there is no need anymore of the regex.

flore77:
Now js-reporters is working also in IE8 along with `babel-polyfill` or `es5-shim` and `es5-sham`.

lib/Data.js
```diff
@@ -34,21 +36,17 @@ export class Suite {
     return this.tests.concat(childSuiteTests)
   }
 
-  get runtime () {
-    var status = this.status
-
-    if (status === 'skipped' || status === undefined) {
+  getRuntime () {
```
jzaefferer: Is there any reason to export this method? If the object isn't expected to change after it was created, we might as well keep this private (with some workaround, since ES6 classes don't have private scope for methods).

Same for `getStatus()`

flore77: Ok, normally I would rename it to `_getStatus()`so that I will notify it is something internal.

jzaefferer:
Besides the one issue, looking good

lib/Data.js
jzaefferer: How about a getStatus() helper function in the private scope of the module, then in the constructor, calling `this.status = getStatus(this)`?

flore77: Fixed in https://github.com/js-reporters/js-reporters/commit/a83b2534bba5b27dc3d368c9aae06e1802bea44b, sorry for this.

PR #84, opened 2016-08-01
Update spec

flore77:
Fixes a part of #30 by updating the spec with the event details type. Also updates the spec based on the discussion in #62, #79 and the changes introduced by #81.

PR #88, opened 2016-08-08
Fix Mocha assertions prop

flore77:
Fixes https://github.com/js-reporters/js-reporters/pull/84#discussion_r73350749.


jzaefferer:
👍

PR #89, opened 2016-08-08
Add examples

### LINKED ISSUE #27: Add examples to README
Add examples to README as discussed in https://github.com/js-reporters/js-reporters/pull/25#issuecomment-130709569.

flore77:
Fixes #27.


flore77:
I guess this example should be enough. By adding 2 examples would be too long, maybe adding to this one also a global test. I am not sure.

For an implementation example, I think if we provide links to exiting implementationa in real tools, would be enough, or even better?

README.md
```diff
@@ -79,6 +79,247 @@ For `testStart` and `testEnd`, the corresponding test object is passed to the re
 
 When `runStart`, `suiteStart` and `testStart` are emitted, the `status`, `runtime` and `errors` attributes are `undefined`.
 
+## Example
+
+Lets take the below testsuite and see how the reporting shoud be, based on the standard presented above. For implementations examples please check [Usage](#usage) and [Integrations](#integrations) below for implementations in real tools.
```
jzaefferer: s/shoud/should

s/implementations/implementation (the plural s is already on "examples")

Let's drop "for implementations in real tools", that's redundant.

README.md
```diff
@@ -79,6 +79,247 @@ For `testStart` and `testEnd`, the corresponding test object is passed to the re
...
+  it('test1', function() {
+    expect(true).to.be.true;
+    expect(true).to.be.false;
+  });
+});
+```
+
+The emitted events and data will be in the following order:
+
+- **runStart**:
```
jzaefferer: Having an example at this level of detail makes sense, but its too long for this document. Let's extract that to a separate file (`example.md`? `docs/example.md`?). Maybe we can include a compressed version here, something like...

```md
* **runStart**: `{ name: undefined, childSuites: [Suite('suite1')] }`
* **suiteStart**: `{ name: 'suite1', childSuites: [Suite('suite2')] }`
...
```

README.md
```diff
@@ -79,6 +79,247 @@ For `testStart` and `testEnd`, the corresponding test object is passed to the re
...
+    passed: 1,
+    failed: 1,
+    skipped: 0,
+    total: 2
+  },
+  runtime: 3
+}
+```
+
+## Usage
```
jzaefferer: How about "Usage of the adapters"?

README.md
```diff
@@ -79,6 +79,247 @@ For `testStart` and `testEnd`, the corresponding test object is passed to the re
...
+ } else if (jasmine) {
+  runner = new JsReporters.JasmineAdapter(jasmine.getEnv());
+ }
+ 
+ // Listen to the same events for any testing framework.
+ runner.on('runStart', function(globalSuite) {
+   ...
+ });
+ 
+ runner.on('testEnd', function(test) {
```
jzaefferer: Let's listen to `testEnd` and print failed assertions and `runEnd` and print a summary. That would be a more useful and practical example.

README.md
```diff
@@ -79,6 +79,247 @@ For `testStart` and `testEnd`, the corresponding test object is passed to the re
...
+   ...
+ });
+ 
+ // Or use one of the built-in reporters.
+ JsReporters.TapReporter.init(runner);
+```
+
+## Integrations
+
+ - [BrowserStack](https://github.com/browserstack/browserstack-runner/blob/master/lib/_patch/reporter.js)
```
jzaefferer: Its "browserstack-runner", BrowserStack is the product, browserstack-runner the tool making use of the product.

flore77: Sorry about this, I keep doing this mistake, but browserstack is shorter than browserstack-runner :laughing:

flore77:
@jzaefferer please review

README.md
```diff
@@ -110,6 +110,53 @@ Based on the discussion in [#79](https://github.com/js-reporters/js-reporters/is
 
 Additional properties (not defined here) can be added to the Assertion object.
 
+## Details
+
+For details please check out the [docs](https://github.com/js-reporters/js-reporters/blob/add-examples/docs) and especially the [example](https://github.com/js-reporters/js-reporters/blob/add-examples/docs/example.md) which presents a testsuite and its reporting data based on the standard presented above.
```
jzaefferer: The paths are wrong, need to remove the extra "add-". Might also be better to use relative links, since absolute links "break" for forks.

README.md
```diff
@@ -110,6 +110,53 @@ Based on the discussion in [#79](https://github.com/js-reporters/js-reporters/is
...
+
+For implementations examples please check [Usage of the adapters](#usage-of-the-adapters) and [Integrations](#integrations).
```
jzaefferer: Remove plural S on "implementations"

README.md
```diff
@@ -110,6 +110,53 @@ Based on the discussion in [#79](https://github.com/js-reporters/js-reporters/is
...
+  runner = new JsReporters.JasmineAdapter(jasmine.getEnv());
+}
+ 
+ // Listen to the same events for any testing framework.
+ runner.on('testEnd', function(test) {
+   console.log('Test %s has errors:', test.fullname.join(' '), test.errors);
+ });
+ 
+ runner.on('runEnd', function(globalSuite) {
+   var testCounts = globalSuite.testCounts;
```
jzaefferer: Drop this or use it

README.md
```diff
@@ -110,6 +110,53 @@ Based on the discussion in [#79](https://github.com/js-reporters/js-reporters/is
...
+   
+   console.log('Build status: %d', globalSuite.status);
```
jzaefferer: s/Build/Test? "Build" seems odd.

README.md
```diff
@@ -110,6 +110,53 @@ Based on the discussion in [#79](https://github.com/js-reporters/js-reporters/is
...
+## Usage of the adapters
+
+Listen to the events and receive the emitted data:
+
+```js
+var runner;
+
+// Attach the adapter.
+if (QUnit) {
+  runner = new JsReporters.QUnitAdapter(QUnit);
```
jzaefferer: I wonder if we should add something like `var runner = JsReporters.guess()`, which looks at globals like this if/else/else does and throws an error when it can't find any framework.

jzaefferer:
The readme should linke to `docs/frameworks.md` somewhere.

README.md
flore77: I am using it below.

README.md
flore77: Yes, sounds good. I will make a separate pr.

README.md
flore77: I was thinking to underline how easily is to get the status for the whole testing, like for continuous integration.

Maybe "Testing status" or "Testsuite status", "Test" sounds a little bit confusing, because you we'll think that is the status only of one test.

README.md
jzaefferer: Oops, right, nevermind

README.md
jzaefferer: "Testsuite status" seems good to me

jzaefferer:
Still missing a direct reference to `docs/frameworks.md` in the README, other looking good.

jzaefferer:
That sounds good!

README.md
```diff
@@ -156,6 +156,17 @@ if (QUnit) {
 
  - [browserstack-runner](https://github.com/browserstack/browserstack-runner/blob/master/lib/_patch/reporter.js) 
 
+## Differences
+
+This section is dedicated to explain the limitations of the adapters in report to the standard.
```
jzaefferer: "in respect"? Instead of "in report"

README.md
```diff
@@ -156,6 +156,17 @@ if (QUnit) {
...
 
+## Differences
+
+This section is dedicated to explain the limitations of the adapters in report to the standard. 
+
+The only limitation is the emitting order, which is not done in source order:
+
+- Jasmine: the emitting order of the tests will be the one from Jasmine
+- Mocha: the emitting order of the tests will be the one from Mocha
+- QUnit: the suites will be emitted in source order, but how a suite was emitted all its tests will be also emitted no matter the source order
```
jzaefferer: I don't understand the part from ",but how", can you rephrase that?

flore77:
@jzaefferer done.

jzaefferer:
Nice, much more clear now. Should be good to land.

PR #90, opened 2016-08-15
Add autoregister function

flore77:
This adds an `autoregister` function in which js-reporters checks for global testing frameworks and registers the respective adapter returning the runner.


lib/helpers.js
```diff
@@ -0,0 +1,21 @@
...
+  var runner
+
+  if (QUnit) {
+    runner = new QUnitAdapter(QUnit)
+  } else if (mocha) {
+    runner = new MochaAdapter(mocha)
+  } else if (jasmine) {
+    runner = new JasmineAdapter(jasmine.getEnv())
+  } else {
+    throw new Error('JsReporters: No testing framework was found')
```
jzaefferer: Could mention the supported frameworks here.

jzaefferer:
Adding a function without tests and not even docs seems fishy. How do we know it works?

flore77:
You are right, I will add some unit tests.

flore77:
@jzaefferer I tried to create some unit tests, but I am not able to mock `QUnitAdapter` functions, because internally in the bundle there is something like:

```js
  var QUnitAdapter = function() {
    ...
  };

  function autoRegister() {
    if (QUnit) {
      var runner = new QUnitAdapter(QUnit);
    }
    ....
  }

  var index = {
    QUnitAdapter: QUnitAdapter
    ...
  };

  return index;
```

Basically I am mocking `index.QUnitAdapter`, but this is not used inside the module and I don't know how to get to the internal `QUnitAdapter` to mock it. Any idea for a workaround ? Or how we should test it?

jzaefferer:
Can you mock `QUnit` and test if any of its methods get called by the adapter?

test/unit/helpers.js
```diff
@@ -0,0 +1,71 @@
...
+
+      JsReporters.autoRegister()
+
+      expect(GLOBAL.QUnit.begin).to.have.been.calledOnce
+
+      delete GLOBAL.QUnit
+    })
+
+    it('should register the MochaAdapter', function () {
+      GLOBAL.QUnit = undefined
```
jzaefferer: Why is this needed?

test/unit/helpers.js
```diff
@@ -0,0 +1,71 @@
...
+        log: dummyFunc,
+        testDone: dummyFunc,
+        done: dummyFunc
+      }
+
+      JsReporters.autoRegister()
+
+      expect(GLOBAL.QUnit.begin).to.have.been.calledOnce
+
+      delete GLOBAL.QUnit
```
jzaefferer: Might be good to move this to an `afterEach`, to prevent this from leaking into another test.

test/unit/helpers.js
```diff
@@ -0,0 +1,71 @@
...
+      GLOBAL.QUnit = undefined
+      GLOBAL.mocha = {
+        reporter: sinon.stub()
+      }
+
+      JsReporters.autoRegister()
+
+      expect(GLOBAL.mocha.reporter).to.have.been.calledOnce
+
+      delete GLOBAL.Jasmine
```
jzaefferer: This looks wrong, should be removing `GLOBAL.mocha` instead

lib/helpers.js
```diff
@@ -0,0 +1,26 @@
...
+  var runner
+
+  if (QUnit) {
+    runner = new QUnitAdapter(QUnit)
+  } else if (mocha) {
+    runner = new MochaAdapter(mocha)
+  } else if (jasmine) {
+    runner = new JasmineAdapter(jasmine.getEnv())
+  } else {
+    throw new Error(`JsReporters: No Qunit, Mocha or Jasmine
```
jzaefferer: How about "Failed to register js-reporters adapater. Supported frameworks are: QUnit, Mocha, Jasmine"?

jzaefferer:
Still missing docs. What's here looks good.

flore77:
@jzaefferer do you agree with https://github.com/js-reporters/js-reporters/pull/90/commits/2c452e11f437e01d5bb8bea6167a3e33c943fc02?

I also added the docs.

When this function will run in the browser? Will work? Shoud we do:
```js
  if (QUnit || window.QUnit) {
    ...
  }
  ...
```

index.js
```diff
@@ -4,7 +4,7 @@ import MochaAdapter from './lib/adapters/MochaAdapter.js'
 import TapReporter from './lib/reporters/TapReporter.js'
 import ConsoleReporter from './lib/reporters/ConsoleReporter.js'
 import {Assertion, Test, Suite} from './lib/Data.js'
-import {autoRegister} from './lib/helpers.js'
```
jzaefferer: I don't understand this change: What was the motivation? This looks worse to me, compared to what it was before.

README.md
```diff
@@ -152,6 +144,12 @@ if (QUnit) {
  JsReporters.TapReporter.init(runner);
 ```
 
+## API
+
+**autoRegister()**
+
+Auto registers one of the existing adapters by checking for existing testing frameworks in the global scope and returns the runner to attach event listeners.
```
jzaefferer: I guess this is the place to also document the `Suite` and `Test` classes?

README.md
```diff
@@ -152,6 +144,12 @@ if (QUnit) {
  JsReporters.TapReporter.init(runner);
 ```
 
+## API
+
+**autoRegister()**
```
jzaefferer: Usually a usage example is pretty much required. In this case, pointing out the example above would be fine.

jzaefferer:
I'm not sure about the browser globals, definitely need to test that.

Maybe we can run one of the test suites in PhantomJS? That's relatively easy to get running on CI.

index.js
flore77: I have done this to be easier to export the whole `helpers`. For example if you want to use some functions to from the `helpers` to can easily import it, not to require every function like:
```js
import {createSuiteStart, createSuiteEnd, createTestStart, createTestEnd} from 'js-reporters/lib/helpers'
```
to be instead:
```js
import helpers from `js-reporters/lib/helpers`
```

flore77: Although every function will be [exported individually](https://github.com/js-reporters/js-reporters/blob/349292be2e8e115b71b476b487897080055a93c4/index.js#L18).

README.md
flore77: I will document them in #92.

flore77:
Which test suite do you mean to run in PhantomJS? And how should I do it, if you can provide some more details. I found [mocha-phantomjs](https://github.com/nathanboktae/mocha-phantomjs) to run mocha tests.

index.js
jzaefferer: You don't need this default export for that. Instead do `import * as helpers from '...'`.

flore77: You are right!

jzaefferer:
That is still on phantomjs 1.x, maybe https://github.com/nathanboktae/mocha-phantomjs-core is a better choice? I can't quite tell what the difference is...

flore77:
@jzaefferer I don't know if that package would work since we have also `chai` and espacially `sinon`. I will have a better look.

Meanwhile I have managed to run the tests in the browser with `Karma` and `Webpack`, which I must admit is pretty difficult to setup. `Sinon` is making a lot of trouble when bundled with `Webpack`, I used the `2.0.0-pre` version which was ok, but this version does not satisfy some peer dependencies, but anyway they seem to work. 

But now i have some problem with `GLOBAL` which is not defined in the browser. I tried to do something like `var gl = GLOBAL || global`, but the tests fail anyway with the error `GLOBAL is undefined`. Why isn't it evaluated to `undefined`? Maybe Karma runs the tests in `strict mode`?

flore77:
I managed to run the `helpers` unit tests also in the browser, I have done some tricks to make it work with sinon. 

It does not work for the `tap-reporter` tests, which are using `fake timers` (more specifically `sinon.test` I guess) and the tricks don't work for this. We can add them after sinon would make a release of it's second version, which works perfectly with webpack.

The tests are failing due to the new release of [Jasmine 2.5.0](https://github.com/jasmine/jasmine/releases), I will investigate.

jzaefferer:
Nice work! Should I review this now, or wait till you resolved the Jasmine issue?

flore77:
No, let me fix the Jasmine thing so we can see that everything is ok. 

I found the issue, it is something from Jasmine, from their node package, I already opened a pr, https://github.com/jasmine/jasmine-npm/pull/89. There should be some changes also in our repo, I will open a pr with the solution and explanations.

flore77:
@jzaefferer please review :smiley:

jzaefferer:
👍

PR #92, opened 2016-08-17
Break Suite and Test into SuiteStart/SuiteEnd and TestStart/TestEnd

### LINKED ISSUE #91: Helpers
While working on the [qunit integration](https://github.com/jquery/qunit/pull/1026), I added js-reporters as a dependency. In my opinion js-reporters should come with some helpers so that it will be easy to adopt the standard.

The scenario that I came through is that when I want to [emit](https://github.com/jquery/qunit/pull/1026/files#diff-7eb52b366866677666470e019283c8eaR251) an object (suite or test) , I don't want to emit the internal object, I want to emit a copy of it and I would like to have a function that would do this for me. Something like:

```js
export function createSuiteStart(suite) {
  return new Suite(
    suite.name,
    suite.fullName.slice(),
    suite.childSuites.map((childSuite) => createSuiteStart(childSuite)),
    suite.tests
  );
}

export function createSuiteEnd(suite) {
  ...
}
```

Also I came into a sort of problem, I am firstly creating the [suites](https://github.com/jquery/qunit/pull/1026/files#diff-7eb52b366866677666470e019283c8eaR61) with an empty array for tests, and then I am adding the [tests](https://github.com/jquery/qunit/pull/1026/files#diff-f6c2ded91c9881cdc31b5bfcabb54c3bR56) to it.

Our Suite implementation sees that there is no test and thinks that the suite has no tests and computes already the `status`, `testCounts` and `runtime`, although I need a `suite start` for which this properties should be be undefined.

If we implement the above functions the emitted objects will be correct (the `createSuiteStart` function will return a Suite with `status` etc. `undefined`). But internally the status of the suite will always be wrong (it will be passed), which I don't find ok.

Now I don't know if I should try to create the suites after I create the tests, or to change something in js-reporters. 

Maybe something like:

```js
class Suite (name, fullName, childSuites, tests, status, testCounts, runtime) {
  ...
  this.status = (arguments.length === 5) ? status : getStatus(this);
  this.testCounts = (arguments.length === 6) ? testCounts : getTestCounts(this);
  ...
}
```

or

```js
class SuiteStart(...) {
  ...
}

class SuiteEnd(...) {
  ...
}
```

flore77:
Fixes #91.

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,41 +63,60 @@ For more background check [#62](https://github.com/js-reporters/js-reporters/iss
 
 ### Event Data
 
-Based on the discussion in [#12](https://github.com/js-reporters/js-reporters/issues/12#issuecomment-120483356), there are two basic data structures: Suites and Tests. A test represents an atomic test/spec/`it()`. A suite contains tests and optionally other suites.
+Based on the discussion in [#12](https://github.com/js-reporters/js-reporters/issues/12#issuecomment-120483356), there are two basic data structures: **Suites** and **Tests**. A test represents an atomic test/spec/`it()`. A suite contains tests and optionally other suites.
 
-For `testStart` and `testEnd`, the corresponding test object is passed to the reporter. The same applies to `suiteStart` and `suiteEnd` where the matching suite object is passed to the reporter. For `runStart` and `runEnd` a "global" suite object is passed to the reporter, this suite wraps all other top-level user defined suites as its child suites, it will be reffered to as `globalSuite`.
+The structures have been divided in two parts, a start part and an end part:
+ - SuiteStart
+ - SuiteEnd
+ - TestStart
+ - TestEnd
+
+For `testStart` and `testEnd`, the corresponding TestStart, respectively TestEnd, object is passed to the reporter. The same applies to `suiteStart` and `suiteEnd` where the matching SuiteStart/SuiteEnd object is passed to the reporter. For `runStart` and `runEnd` a "global" suite object is passed to the reporter, this suite wraps all other top-level user defined suites as its child suites, it will be reffered to as `globalSuite`.
 
 The data structures are defined as follows:
 
-- **Suite**: `Object` - A suite is a collection of tests and potentially other suites.
+- **SuiteStart**: `Object` - A SuiteStart is a collection of tests-starts and potentially other suites-starts, emitted before the suite have been executed, which means before executing any of its tests and child suites.
  - **name**: `String|undefined` - name of the suite, will be `undefined` only for the `globalSuite`.
  - **fullname**: `Array` - array of strings containing the name of the suite and the names of all its suites ancestors.
  - **tests**: `Array` - array containing all tests that directly belong to the suite (but not to a child suite).
  - **childSuites**: `Array` - array with all direct subsuites.
- - **status**: `String|undefined` - summarized status of the suite, it will be `undefined` on the `suiteStart` and `runStart` event.
+ - **testCounts**: `Object` - contains the total number of tests in the suite, including the tests of the child suites.
+    - **total**: `Number` - total number of tests
+
+- **SuiteEnd**: `Object` - A SuiteEnd is a collection of tests-ends and potentially other suites-ends, emitted after the suite has been executed, which means that all its tests and child suites have been also executed.
+ - **name**: `String|undefined` - name of the suite, will be `undefined` only for the `globalSuite`.
+ - **fullname**: `Array` - array of strings containing the name of the suite and the names of all its suites ancestors.
+ - **tests**: `Array` - array containing all tests that directly belong to the suite (but not to a child suite).
+ - **childSuites**: `Array` - array with all direct subsuites.
+ - **status**: `String` - summarized status of the suite.
     - `failed`, if at least one test in the suite or in its child suites has failed.
     - `skipped`, if all tests in the suite and in its child suites are skipped (and there is at least one skipped test).
     - `passed`, if there is at least one passed test in the suite or in its child suites and all other tests are skipped or if there are no tests in the suite.
  - **testCounts**: `Object` - contains how many tests have passed, failed etc. including the tests of child suites.
-    - **passed**: `Number|undefined` - number of passed tests, it will be `undefined` for the `runStart` and `suiteStart` event.
-    - **failed**: `Number|undefined` - number of failed tests, it will be `undefined` for the `runStart` and `suiteStart` event.
-    - **skipped**: `Number|undefined` - number of skipped tests, it will be `undefined` for the `runStart` and `suiteStart` event.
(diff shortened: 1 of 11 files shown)


flore77:
@jzaefferer what do you think so far? Also by adding helper functions like `createSuiteStart` etc., we will simplify the code in the adapters, more exactly in [JasmineAdatper](https://github.com/js-reporters/js-reporters/blob/0cddf82153db8ea71567dd5d8387e9cc0205ff19/lib/adapters/JasmineAdapter.js#L26-L58) and [QunitAdapter](https://github.com/js-reporters/js-reporters/blob/master/lib/adapters/QUnitAdapter.js#L136-L161).

jzaefferer:
Looks good! Can you update the spec as well?

jzaefferer:
Spec update also looks good!

flore77:
I don't know what to do exactly with the `errors` and `assertions` [arrays](https://github.com/js-reporters/js-reporters/pull/92/files#diff-27a1c86133de4310373decfdad200a91R53), there will be the same arrays also in the new created object, and `slice` wouldn't work because there objects, `Assertions`. Would it be fine to leave them like this ?

jzaefferer:
> there will be the same arrays also in the new created object

Could you clarify this? I don't follow.

flore77:
There will be no copies of those arrays, currently it will be the same array.

```js
var errors = [];
var t = new Test():

t.errors = errors;
```

`t.errors` will point to the same array, this can be changed through:
```js
t.errors = errors.slice();
```
It will be another array, but sine in the array are objects, in the new created array will be the same objects and if some modifies `t.errors` it will also reflect in the original `errors` array.

And I am not sure how to handle this.

jzaefferer:
Okay, though I still don't understand the context where this is a problem. As long as `TestStart` doesn't have assertions and errors, the array should exist on `TestEnd`, right?

flore77:
Yes, there exists only on `TestEnd`. The problem would be if someone changes something, it would reflect also in our internal data structures. So the `createTestEnd` function does not create a completely new object and I don't know if this is fine or no.

jzaefferer:
That seems fine to me.

flore77:
Alright, then please review :smiley:

README.md
```diff
@@ -63,41 +63,60 @@ For more background check [#62](https://github.com/js-reporters/js-reporters/iss
...
+ - SuiteEnd
+ - TestStart
+ - TestEnd
+
+For `testStart` and `testEnd`, the corresponding TestStart, respectively TestEnd, object is passed to the reporter. The same applies to `suiteStart` and `suiteEnd` where the matching SuiteStart/SuiteEnd object is passed to the reporter. For `runStart` and `runEnd` a "global" suite object is passed to the reporter, this suite wraps all other top-level user defined suites as its child suites, it will be reffered to as `globalSuite`.
 
 The data structures are defined as follows:
 
-- **Suite**: `Object` - A suite is a collection of tests and potentially other suites.
+- **SuiteStart**: `Object` - A SuiteStart is a collection of tests-starts and potentially other suites-starts, emitted before the suite have been executed, which means before executing any of its tests and child suites.
```
jzaefferer: How about "test-starts" and "suite-starts"?

README.md
```diff
@@ -63,41 +63,60 @@ For more background check [#62](https://github.com/js-reporters/js-reporters/iss
...
  - **name**: `String|undefined` - name of the suite, will be `undefined` only for the `globalSuite`.
  - **fullname**: `Array` - array of strings containing the name of the suite and the names of all its suites ancestors.
  - **tests**: `Array` - array containing all tests that directly belong to the suite (but not to a child suite).
  - **childSuites**: `Array` - array with all direct subsuites.
- - **status**: `String|undefined` - summarized status of the suite, it will be `undefined` on the `suiteStart` and `runStart` event.
+ - **testCounts**: `Object` - contains the total number of tests in the suite, including the tests of the child suites.
+    - **total**: `Number` - total number of tests
+
+- **SuiteEnd**: `Object` - A SuiteEnd is a collection of tests-ends and potentially other suites-ends, emitted after the suite has been executed, which means that all its tests and child suites have been also executed.
+ - **name**: `String|undefined` - name of the suite, will be `undefined` only for the `globalSuite`.
```
jzaefferer: How about removing the duplication here by saying something like "In addition to the properties of `SuiteStart`, `SuiteEnd` also has the following properties:"?

README.md
```diff
@@ -63,41 +63,60 @@ For more background check [#62](https://github.com/js-reporters/js-reporters/iss
...
     - `passed`, if all assertions have passed.
     - `failed`, if at least one assertion has failed.
     - `skipped`, if the test is disabled and wasn't executed.
- - **runtime**: `Number|undefined` - execution time in milliseconds, it will be `undefined` for the `testStart` event.
- - **errors**: `Array|undefined` - array containing all errors, i.e failed Assertions, it will be `undefined` for the `testStart` event. It will contain at least one error for failed statuses and it will be empty for statuses other than failed.
- - **assertions**: `Array|undefined` - array of Assertions containing all assertions passed and failed, for a skipped test there will be an empty array, it will be `undefined` for the `testStart` event. Frameworks that don't track passed assertions can always provide an empty array for passed tests. In that case, for failed tests this should match the errors property.
+ - **runtime**: `Number` - execution time in milliseconds.
+ - **errors**: `Array|undefined` - array containing all errors, i.e failed Assertions. It will contain at least one error for failed statuses and it will be empty for statuses other than failed.
```
jzaefferer: Is the `|undefined` still valid? Seems like that should be dropped.

flore77:
Finished changes.

README.md
```diff
@@ -75,19 +75,15 @@ For `testStart` and `testEnd`, the corresponding TestStart, respectively TestEnd
...
  - **childSuites**: `Array` - array with all direct subsuites.
  - **testCounts**: `Object` - contains the total number of tests in the suite, including the tests of the child suites.
     - **total**: `Number` - total number of tests
 
-- **SuiteEnd**: `Object` - A SuiteEnd is a collection of tests-ends and potentially other suites-ends, emitted after the suite has been executed, which means that all its tests and child suites have been also executed.
- - **name**: `String|undefined` - name of the suite, will be `undefined` only for the `globalSuite`.
- - **fullname**: `Array` - array of strings containing the name of the suite and the names of all its suites ancestors.
- - **tests**: `Array` - array containing all tests that directly belong to the suite (but not to a child suite).
- - **childSuites**: `Array` - array with all direct subsuites.
+- **SuiteEnd**: `Object` - A SuiteEnd is a collection of tests-ends and potentially other suites-ends, emitted after the suite has been executed, which means that all its tests and child suites have been also executed. **In addition** to the properties of `SuiteStart`, `SuiteEnd` also has the following properties:
```
jzaefferer: Let's fix these tests-ends and suites-ends, too

PR #93, opened 2016-09-01
Fix reporter adding in Jasmine adapter

flore77:
This solves the adding of the jasmine adapter in a node env.

We cannot pass the `jasmine.env` to the reporter, we must pass in the `jasmine` object, which is a NodeJS helper over `jasmine-core`. We must use the `jasmine.addReporter` and **not** directly `jasmine.env.addReporter`, because then it will enter on this [branch](https://github.com/jasmine/jasmine-npm/blob/164b799c6e531f29b23722cf6975dfe1400ac7b6/lib/jasmine.js#L163) and jasmine will exit and kill the process. Using the `jasmine.addReporter` has some internal logic that holds from exiting.

Now the adapter has some extra logic to differentiate btw a node env and a browser env. Unfortunately we cannot test the browser part, because currently the tests are running only in a node env.

### DIFF
--- a/lib/adapters/JasmineAdapter.js
+++ b/lib/adapters/JasmineAdapter.js
@@ -10,17 +10,27 @@ export default class JasmineAdapter extends EventEmitter {
     super()
 
     this.jasmine = jasmine
+    // NodeJS/browser
+    this.env = jasmine.env || jasmine.getEnv()
     this.suites = {}
     this.tests = {}
 
-    jasmine.addReporter({
+    var reporter = {
       jasmineStarted: this.onJasmineStarted.bind(this),
       specDone: this.onSpecDone.bind(this),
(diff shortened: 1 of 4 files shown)


jzaefferer:
Looks good, except for failing tests.

Seems like this will need an update as well: https://github.com/js-reporters/js-reporters/pull/90/files#diff-27a1c86133de4310373decfdad200a91R19

flore77:
> Seems like this will need an update as well: https://github.com/js-reporters/js-reporters/pull/90/files#diff-27a1c86133de4310373decfdad200a91R19

This will be solved in #90, it will be rebased on the master after this pr will be merged.

Good to land? :smiley:

flore77:
I set jasmine to the last working versions, after a version with the fix will be published.

PR #94, opened 2016-12-10
Specify data for Todo tests

trentmwillis:
This proposes adding a new status of `todo` for both Suites and Tests. It is similar to the `skipped` status in that it should be counted separately from `passed` tests, but is also different because it is possible that `todo` tests can also fail.

`todo` should represent tests where there is some amount of incomplete work, meaning that there is at least one failing assertion in the test. These assertions should still be reported as normal, but should not cause a `todo` test to fail. A `todo` only fails when all assertions have passed. In that case, the status should be reported as `failed` so that reporters can properly notify that the feature is now implemented.

_Original discussion occurred in https://github.com/qunitjs/qunit/pull/1006._

cc @leobalter

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,7 +46,7 @@ Based on the discussion in [#1](https://github.com/js-reporters/js-reporters/iss
  - `suiteEnd`:  Triggered at the end of each group of tests within a testsuite.
  - `runEnd`:  Indicates the end of a testsuite, triggered just once.
 
- 
+
 #### Selection Criteria
 
 The criteria for picking these event names included:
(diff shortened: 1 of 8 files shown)


jzaefferer:
The spec change looks alright. It would be nice to get the adapters updated as well, which would also tell us if any other framework has a similar feature. Considering that this spec still has almost zero traction, it might be a waste of time though.

trentmwillis:
As far as I can tell, no framework actually implements "todo" tests. Mocha and Jasmine have "pending" tests, but those are essentially the same as "skipped" tests (just no callback defined). Thus, I don't think updating the adapters at this point in time makes much sense (since we don't know what we're adapting).

trentmwillis:
Update: we're about to land `todo` tests in QUnit and have work in progress to land js-reporters shortly after that. In which case, we will be using the approach documented in this PR unless any objections get raised.

jzaefferer:
Could you add a note to the spec that `todo` is something new in QUnit, along with a link to the docs or something?

trentmwillis:
@jzaefferer I've updated this with more documentation for `todo`. I also added the data to the TapReporter, the Data helpers, and fixed CI.

That said, I still need to update the QUnitAdapter now, but I'd prefer to do that in a follow up PR to avoid this one getting to complex (though it might already be there).

PR #99, opened 2017-03-30
Readme: Use canonical path to QUnit repository

platinumazure:
GitHub will redirect jquery/qunit URLs to qunitjs/qunit, but it's good to just use the correct URLs going forward.


leobalter:
You may also change [`docs/frameworks.md` at line 74](https://github.com/js-reporters/js-reporters/blame/master/docs/frameworks.md#L74)

PR #102, opened 2017-05-17
Reporter: Print "actual:", "expected:" even if they are undefined

mo:
If "foo" was expected but the actual value was undefined, then print:

  actual: undefined
  expected: "foo"

If undefined was expected but "foo" was returned, then we print:

  actual: "foo"
  expected: undefined

NOTE: When "actual" and "expected" are both undefined, the error
being logged is not one comparing two values, for example:

  message: "Too many calls to the `assert.async` callback"
  severity: failed

### DIFF
--- a/lib/reporters/TapReporter.js
+++ b/lib/reporters/TapReporter.js
@@ -46,12 +46,14 @@ export default class TapReporter {
     console.log(`  message: "${error.message || 'failed'}"`)
     console.log(`  severity: ${severity || 'failed'}`)
 
-    if (error.actual) {
-      console.log(`  actual: ${JSON.stringify(error.actual, null, 2)}`)
+    if (error.hasOwnProperty('actual')) {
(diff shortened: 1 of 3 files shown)


lib/reporters/TapReporter.js
```diff
@@ -46,12 +46,14 @@ export default class TapReporter {
     console.log(`  message: "${error.message || 'failed'}"`)
     console.log(`  severity: ${severity || 'failed'}`)
 
-    if (error.actual) {
-      console.log(`  actual: ${JSON.stringify(error.actual, null, 2)}`)
+    if (error.hasOwnProperty('actual')) {
+      var actualStr = error.actual ? JSON.stringify(error.actual, null, 2) : 'undefined'
```
trentmwillis: This check should be stricter, otherwise we'll get `undefined` for any "falsy" value.

mo: fixed now

PR #107, opened 2018-05-19
Reporter: TapReporter align actual with expected.

rwjblue:
When comparing two strings, it is significantly easier to spot the difference when they are aligned. This change pads the label for `actual` with a couple of spaces so that the actual value is printed at the same indentation as the expected value. Previously the expected value was right shifted by 2 characters...

Before:

2018-05-18_23-31-32_65aka_tmux image

After:

2018-05-18_23-29-33_ogp95_tmux image


rwjblue:
Thoughts?

jzaefferer:
This certainly looks like an improvement!

PR #115, opened 2019-11-20
Reporter: Escape backslashes in YAML output

### LINKED ISSUE #114: Backslashes should be escaped in YAML strings.
If an error message contains backslashes, YAML output looks like this:

```yaml
  ---
  message: "Test failed\"
  severity: failed
  ...
```

which is not valid since the backslash now looks like it's escaping the closing double quote. Instead it should be:

```yaml
  ---
  message: "Test failed\\"
  severity: failed
  ...
```

jeberger:
Fixes #114

### DIFF
--- a/lib/reporters/TapReporter.js
+++ b/lib/reporters/TapReporter.js
@@ -42,7 +42,12 @@ export default class TapReporter {
   }
 
   quote (str) {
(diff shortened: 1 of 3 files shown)


Krinkle:
I believe this has since been fixed by https://github.com/js-reporters/js-reporters/pull/110.

Issue #105, opened 2017-12-18
`todo` property on assertions feels incorrect


keithamus:
In the current readme, assertion objects have a `todo` parameter - indicating whether or not this was part of a todo test. This feels invalid for a few reasons:

 - As a "separation of concerns" principle, the assertion should not know about its parent state.
 - More concretely to the above bullet point, assertion libraries such as [chai](https://github.com/chaijs/chai) _cannot_ know about the test state.
 - This doesn't seem to cover all states in a test, e.g. `skipped` isn't captured info in an assertion error. If skipped isn't, why is todo?
 - Existence of the property on the assertion object implies some assertions (within the same test suite) can be todo while others cannot. 
 - This violates the principle of single source of truth. What happens if my test object says its state is `"todo"` but my assertion object says `todo: false`? Which one do I believe?

My suggestion is to remove the `todo` property from the assertion object, and have it only exist as a state on the test object.

trentmwillis:
This reasoning seems sound. It was added to assertions data originally because the first implementation within QUnit required it to be present due to the way some of the test reporters work. That said, it should not be a *required* field due to the reasons you list above.

Krinkle:
Proposing we make this optional (or simply remove) in the next major release.

I'm interested to know what makes it hard to track in Chai. I completely agree on the separation of concerns and that Chai shouldn't have to track this state. I'm just curious as to why it would be difficult, as that might lead us to uncover a deeper compat issue or difference to be aware of and for us to document internally.

In particular, because consumers of js-reporters do need to know this information about each high-level test wrapper and will generally have that information available indirectly for each assertion, so that they could e.g. generate a nested or flat results lists as they wish where any test-information is repeated (e.g. by prefixing the suite names and test names, and their skipped/todo status etc.). Would that not be possible either?

@keithamus Could you elaborate on why Chai can't know about this state?

Krinkle:
I propose that we simply remove it entirely from assertion object. The spec allows additonal non-standard properties to exist, so this would be a backwards-compatible change to make.

Even on QUnit it is always decided per-test, it cannot be set per-assertion.

One thing to keep in mind is how reporters need to read the "todo" status. I recently rediscovered https://github.com/karma-runner/karma-qunit/issues/111 which suggests at least in the past there was confusing over how a reporter is meant to consume "todo" events - where Karma currently wrongly fails the build due to perceiving the todo failures as real failures.

keithamus:
[...]

PR #119, opened 2020-09-24
Spec: Remove "todo" from Assertion event data

Krinkle:
* This is redundant with the information already available in
  in the TestEnd event data.

* We don't know of any testing frameworks that support flipping
  the "todo" status on a per-assertion basis, nor does that
  seem likely in the future as "Todo test" is defined as a test
  with one or more expected failures.

* Presence of this in the spec was a barrier to entry as it means
  common assertion exception objects would require being cloned,
  mutated, or otherwise mapped to set accordingly.

Fixes https://github.com/js-reporters/js-reporters/issues/105.

### DIFF
--- a/lib/Data.js
+++ b/lib/Data.js
@@ -79,15 +79,13 @@ class Assertion {
    * @param {*} expected
    * @param {String} message
    * @param {String|undefined} stack
-   * @param {Boolean} todo
    */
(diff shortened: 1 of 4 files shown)


keithamus (approved):
:+1:

Issue #126, opened 2020-11-12
Allow for tests nested within tests


Krinkle:
* Allow tests to directly contain both assertions and other tests.

Per https://github.com/js-reporters/js-reporters/issues/117#issuecomment-713903902 this is required for node-tap and tape. This does limit a bit how reporters can visualise and lay out the data, but seems worth doing. Especially as it aligns us closer with TAP.

For HTML reporters like QUnit that provide a collapsible list of assertions for each, they may need to buffer each test and then render the list of assertions first, followed by the sub tests. This is a small price to pay for widening up the surface of test frameworks and reporters that can participate. It also wouldn't negatively affect any reporters that exist today since those are currently specific to frameworks frameworks that would never excercise that need for buffering, so it's only a win-win to allow the reporter to be user more widely.

* Consider phasing out the "Suite" concept. Most likely obsolete after this.

Krinkle:
Requiring consumers to support assertions and nested tests under the same concept does the choices a reporter can use for their visual or textual layout. For assertions to appear between tests, and tests between assertions.

I realize this may insignificant from a TAP perspective. From what I've seen in the TAP community, frameworks are generally exception-based, runners stop after the first error, and reporters only communicate tests and failed assertions. By not rendering multiple errors at different levels, and by not making passed assertions accessible, this means one conveniently sidestep this issue. QUnit is not exception-based, may provide multiple errors from a single run, and does make its passed assertions accessible via a click or flag. Below is an axample of its HTML report, with two of the tests expanded to show its assertions:

<img alt="Screenshot" src="https://user-images.githubusercontent.com/156867/98876214-b1effa00-2475-11eb-9e59-a5284e97414c.png" width="400">

I suppose individual consumers wishing to render things this way could workaround this limitation by buffering any nested tests, and rendering the direct assertions first. This would slightly misrepresent the true data, but seems and is no worse than what they can do today. It just means they can be used by a wider range of test frameworks.

Krinkle:
I'm running into a few issues when trying to merge the "suite" and "test" concepts.

The simplest approach, I thought, would be to simply convert suites to tests and call it a day. However, this is proving to be more difficult than I thought. Our TAP reporter, ignores "suite" events. It only acknowledges suite as a concept by means of prefixing the test names (which are provided as part of the test name, so technically it's entirely unaware of suites).

Before events:

* runStart: (total: 2)
* suiteStart: "some suite"
* testStart: "test A" (parent: "some suite")
[...]

PR #129, opened 2021-01-22
Spec: Combine the "Test" and "Suite" concepts

Krinkle:
### Status quo

The TAP 13 specification does not standardise a way of describing parent-child relationships between tests, nor does it standardise how to group tests.

Yet, all major test frameworks have a way to group tests (e.g. QUnit module, and Mocha suite) and/or allow nesting tests inside of other tests (like tape, and node-tap). While the CRI draft provided a way to group tests, it did not accomodate Tap. They would either need to flatten the tests with a separator symbol in the test name, or to create an implied "Suite" for every test that has non-zero children and then come up with an ad-hoc naming scheme for it.

Note that the TAP 13 reporter we ship, even after this change, still ends up flattening the tests by defaut using the greater than `>` symbol, but at least the event model itself recognises the relationships so that other output formats can make use of it, and in the future TAP 14 hopefully will recognise it as well, which we can then make use of.

Ref https://github.com/TestAnything/testanything.github.io/pull/36.

### Summary of changes

See the diff of `test/integration/reference-data.js` for the concrete changes this makes to the consumable events.

- Remove `suiteStart` and `suiteEnd` events.

  Instead, the spec now says that tests are permitted to have children.

  The link from child to parent remains the same as before, using the `fullName` field which is now a stack of test names. Previously, it was a stack of suite names with a test name at the end.

- Remove all "downward" links from parent to child. Tests don't describe their children upfront in detail, and neither does `runStart`. This was information was very repetitive and tedious to satisy for implementors, and encouraged or required inefficient use of memory.

  I do recognise that a common use case might be to generate a single output file or stream where real-time updates are not needed, in which case you may want a convenient tree that is ready to traverse without needing to listen for async events and put it together. For this purpose, I have added a built-in reporter that simply listens to the new events and outputs a "summary" event with an object that is similar to the old "runEnd" event object where the entire run is described in a single large object.

- New "SummaryReporter" for simple use cases of non-realtime traversing of single structure after the test has completed.

### Caveats

- A test with the "failed" status is no longer expected to always have an error directly associated with it.

  Now that tests aggregate into other tests rather than into suites, this means tests that merely have other tests as children do still have to send a full testEnd event, and thus an `errors` and `assertions` array.

  I considered specifying that errors have to propagate but this seemed messy and could lead to duplicate diagnostic output in  reporters, as well ambiguity or uncertainty over where errors originated.

- A suite containing only "skipped" tests now aggregates as "passed" instead of "skipped". Given we can't know whether a suite is its own test with its own assertions, we also can't assume that if a test parent has only "skipped" children that the parent was also skipped.

  This applies to our built-in adapters, but individual frameworks, if they know that a suite was skipped in its entirety, can of course still set the status of parents however they see fit.

- Graphical reporters (such as QUnit and Mocha's HTML reporters) may no longer assume that a test parent has either assertions/errors or other tests. A test parent can now have both its own assertions/errors, as well as other tests beneath it.

  This restricts the freedom and possibilities for visualisation. My recommendation is that, if a visual reporter wants to keep using different visual shapes for "group of assertions" and "group of tests", that they buffer the information internally such that they can first render all the tests's own assertions, and then render the children, even if they originally ran interleaved and/or the other way around.

  Ref https://github.com/js-reporters/js-reporters/issues/126.

- The "Console" reporter that comes with js-reporter now no longer uses `console.group()` for collapsing nested tests.

### Misc

- Add definitions for the "Adapter" and "Producer" terms.

- Use terms "producer" and "reporter" consistently, instead of "framework", "runner", or "adapter".

- Remove mention that the spec is for reporting information from "JavaScript test frameworks". CRI can be used to report information about any kind of test that can be represented in CRI's event model, including linting and end-to-end tests for JS programs, as well as non-JS programs. It describes a JS interface for reporters, but the information can come from anywhere.

  This further solifies that CRI is not meant to be used for "hooking" into a framework, and sets no expectation about timing or run-time environment being shared with whatever is executing tests in some form or another. This was already the intent originally, since it could be used to report information from other processes or from a cloud-based test runner like BrowserStack, but this removes any remaining confusion or doubt there may have been.

Fixes https://github.com/js-reporters/js-reporters/issues/126.

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,26 +52,25 @@ Would _you_ be interested in discussing this with us further? Please join in!
 Listen to the events and receive the emitted data:
 
 ```js
-// Attach one of the exiting adapters.
+// Use automatic discovery of the framework adapter.
 const runner = JsReporters.autoRegister();
 
-// Listen to the same events for any testing framework.
-runner.on('testEnd', function (test) {
-  console.log('Test %s has errors:', test.fullname.join(' '), test.errors);
+// Listen to standard events, from any testing framework.
+runner.on('testEnd', (test) => {
+  console.log('Test %s has errors:', test.fullName.join(' '), test.errors);
 });
 
-runner.on('runEnd', function (globalSuite) {
-  const testCounts = globalSuite.testCounts;
+runner.on('runEnd', (run) => {
+  const counts = run.counts;
 
-  console.log('Testsuite status: %s', globalSuite.status);
-
+  console.log('Testsuite status: %s', run.status);
   console.log('Total %d tests: %d passed, %d failed, %d skipped',
-    testCounts.total,
-    testCounts.passed,
-    testCounts.failed,
-    testCounts.skipped);
-
-  console.log('Total duration: %d', globalSuite.runtime);
+    counts.total,
+    counts.passed,
+    counts.failed,
+    counts.skipped
+  );
+  console.log('Total duration: %d', run.runtime);
 });
 
 // Or use one of the built-in reporters.
--- a/index.js
+++ b/index.js
@@ -6,9 +6,6 @@ const TapReporter = require('./lib/reporters/TapReporter.js');
 const ConsoleReporter = require('./lib/reporters/ConsoleReporter.js');
 const SummaryReporter = require('./lib/reporters/SummaryReporter.js');
 const {
-  collectSuiteStartData,
-  collectSuiteEndData,
-  createSuiteStart,
   createTestStart
 } = require('./lib/helpers.js');
 const { autoRegister } = require('./lib/auto.js');
@@ -21,9 +18,6 @@ module.exports = {
   ConsoleReporter,
   SummaryReporter,
   EventEmitter,
-  collectSuiteStartData,
-  collectSuiteEndData,
-  createSuiteStart,
   createTestStart,
   autoRegister
 };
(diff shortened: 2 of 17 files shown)


Krinkle:
For a quick overview, see [diff of the test fixture: reference-data.js](https://github.com/js-reporters/js-reporters/pull/129/commits/cf639965fe07170d829b1e25b46cdf93e70b712d#diff-b19c11857c7dd061a5ca9c396367f89074dcd5441600c9c0869b6ffd9d997719).

/cc @isaacs Could use a second pair of eyes on this. The effective difference is quite straight-forward but there are a number of caveats I outlined above. It'd be great if you could confirm that the listed caveats are indeed accepted status quo with node-tap and TAP, or whether there's perhaps another way that I missed. Thanks!

Krinkle:
For a quick overview, see [diff of the test fixture: reference-data.js](https://github.com/js-reporters/js-reporters/pull/129/commits/cf639965fe07170d829b1e25b46cdf93e70b712d#diff-b19c11857c7dd061a5ca9c396367f89074dcd5441600c9c0869b6ffd9d997719) and [summary-reporter.js](https://github.com/js-reporters/js-reporters/pull/129/commits/cf639965fe07170d829b1e25b46cdf93e70b712d#diff-ac4b5a18130f77f8638336f8a816d08b1ab353278a720537c6b260b9868205abR129).

/cc @isaacs @ljharb
Could use a second pair of eyes on this. The effective difference is quite straight-forward but there are a number of caveats I outlined above. It'd be great if you could confirm that the listed caveats are indeed accepted status quo with node-tap and TAP, or whether there's perhaps another way that I missed. Thanks!

PR #130, opened 2021-02-14
Reporter: handle objects with cycles

### LINKED ISSUE #104: TapReporter needs circular ref filtering on JSON.stringify

zackthehuman:
Updating `TapReporter` with the ability to handle objects with circular
references. This is needed for proper stringification of actual and
expected values that contain cycles.

Fixes #104

### DIFF
--- a/lib/reporters/TapReporter.js
+++ b/lib/reporters/TapReporter.js
@@ -30,6 +30,9 @@ const hasOwn = Object.hasOwnProperty;
  * - bare unquoted text, for simple one-line strings.
  * - JSON (quoted text), for complex one-line strings.
  * - YAML Block, for complex multi-line strings.
+ *
+ * Objects with cyclical references will be stringifed as
+ * "[Circular]" as they cannot otherwise be represented.
  */
 function prettyYamlValue (value, indent = 4) {
   if (value === undefined) {
@@ -116,7 +119,46 @@ function prettyYamlValue (value, indent = 4) {
   }
 
   // Handle null, boolean, array, and object
-  return JSON.stringify(value, null, 2);
+  return JSON.stringify(decycledShallowClone(value), null, 2);
+}
+
+/**
+ * Creates a shallow clone of an object where cycles have
+ * been replaced with "[Circular]".
+ */
+function decycledShallowClone (object, ancestors = []) {
+  if (ancestors.indexOf(object) !== -1) {
+    return '[Circular]';
+  }
+
+  let clone;
+
+  const type = Object.prototype.toString
+    .call(object)
+    .replace(/^\[.+\s(.+?)]$/, '$1')
+    .toLowerCase();
+
+  switch (type) {
+    case 'array':
+      ancestors.push(object);
+      clone = object.map(function (element) {
+        return decycledShallowClone(element, ancestors);
+      });
+      ancestors.pop();
+      break;
+    case 'object':
+      ancestors.push(object);
+      clone = {};
+      Object.keys(object).forEach(function (key) {
+        clone[key] = decycledShallowClone(object[key], ancestors);
+      });
+      ancestors.pop();
+      break;
+    default:
+      clone = object;
+  }
+
+  return clone;
 }
 
 module.exports = class TapReporter {
--- a/test/fixtures/unit.js
+++ b/test/fixtures/unit.js
@@ -11,6 +11,39 @@ function copyErrors (testEnd) {
   return testEnd;
 }
 
+/**
+ * Creates an object that has a cyclical reference.
+ */
+function createCyclical () {
+  const cyclical = { a: 'example' };
+  cyclical.cycle = cyclical;
+  return cyclical;
+}
+
+/**
+ * Creates an object that has a cyclical reference in a subobject.
+ */
+function createSubobjectCyclical () {
+  const cyclical = { a: 'example', sub: {} };
+  cyclical.sub.cycle = cyclical;
+  return cyclical;
+}
+
+/**
+ * Creates an object that references another object more
+ * than once in an acyclical way.
+ */
+function createDuplicateAcyclical () {
+  const duplicate = {
+    example: 'value'
+  };
+  return {
+    a: duplicate,
+    b: duplicate,
+    c: 'unique'
+  };
+}
+
 module.exports = {
   passingTestStart: {
     name: 'pass',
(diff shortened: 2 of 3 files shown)


lib/reporters/TapReporter.js
```diff
@@ -116,7 +119,30 @@ function prettyYamlValue (value, indent = 4) {
...
+  return JSON.stringify(value, createCycleSafeReplacer(), 2);
+}
+
+/**
+ * Creates a replacer function to use with JSON.stringify that
+ * safely handles cyclical references. When a cycle is detected
+ * it will be represented as "[Circular]".
+ */
+function createCycleSafeReplacer() {
+  const seen = new Set();
```
ljharb, line 131:
```suggestion
  const seen = new WeakSet();
```

lib/reporters/TapReporter.js
```diff
@@ -116,7 +119,30 @@ function prettyYamlValue (value, indent = 4) {
...
+
+  return function cycleSafeReplacer(_key, value) {
+    let result = value;
+
+    if (typeof value === 'object' && value !== null) {
+      if (seen.has(value)) {
+        return '[Circular]';
```
ljharb, line 138: node’s util.inspect tracks specific refs here, so you know what it’s circular *of* - it might be worth doing that here.

zackthehuman:
I need to crunch on this a little bit more. The existing approach can have false positives as long as the same object appears more than once (even when it's not cyclical). I also have to fix the linting errors.

lib/reporters/TapReporter.js
zackthehuman: This is indeed a neat feature of `inspect`. One question is whether this would be confusing seeing it in the actual/expected output. To label the refs we need two parts:

1. To label the object being referenced as a `<ref *n>`
2. To use the correct index when labeling cycles `[Circular *n]`

Labeling the cycles should be pretty easy, but I'm a little worried about changing the structure of the stringified object too much. 

There are some other enhancements to this stringification we need to consider, for example, `BigInt`. I'd rather follow up on enhancements separately and landing this fix first.

Thoughts?

lib/reporters/TapReporter.js
zackthehuman: I made this change but then later didn't need it since I switched to using a shallow clone and the cloning is stack-based.

lib/reporters/TapReporter.js
ljharb: Right - normal JSON.stringify isn't really an option, I think, at least not without a stringifier function also provided.

zackthehuman:
Circling back on this: I think there are many improvements we can make to the serialization outside of handling ref cycles. It would be worth having an issue dedicated to those improvements so that we can figure out what "ideal" serialization looks like.

In the short term, it would help me (and others) to land this fix. I've squashed/rebased my changes, so it should be in a reviewable state. @Krinkle if you have any time would you mind taking a look?

lib/reporters/TapReporter.js
```diff
@@ -116,7 +119,46 @@ function prettyYamlValue (value, indent = 4) {
...
+    .replace(/^\[.+\s(.+?)]$/, '$1')
+    .toLowerCase();
+
+  switch (type) {
+    case 'array':
+      ancestors.push(object);
+      clone = object.map(function (element) {
+        return decycledShallowClone(element, ancestors);
+      });
+      ancestors.pop(object);
```
Krinkle, line 147: `Array.pop()` does not take an argument.

lib/reporters/TapReporter.js
```diff
@@ -116,7 +119,46 @@ function prettyYamlValue (value, indent = 4) {
...
-  return JSON.stringify(value, null, 2);
+  return JSON.stringify(decycledShallowClone(value), null, 2);
+}
+
+/**
+ * Creates a shallow clone of an object where cycles have
+ * been replaced with "[Circular]".
+ */
+function decycledShallowClone (object, ancestors = []) {
+  if (ancestors.includes(object)) {
```
Krinkle, line 130: Use `Array#indexOf() !== -1` here for browser compat as I'd rather not ship a polyfill with this library, and given its low-level use in test runners it's not feasible for end-users to ship a polyfil either as they generally can't or wouldn't be expected to load something before this.

lib/reporters/TapReporter.js
```diff
@@ -116,7 +119,46 @@ function prettyYamlValue (value, indent = 4) {
...
+      });
+      ancestors.pop(object);
+      break;
+    case 'object':
+      ancestors.push(object);
+      clone = {};
+      Object.keys(object).forEach(function (key) {
+        clone[key] = decycledShallowClone(object[key], ancestors);
+      });
+      ancestors.pop(object);
```
Krinkle, line 155: (idem)

test/fixtures/unit.js
```diff
@@ -11,6 +11,27 @@ function copyErrors (testEnd) {
...
+  cyclical.cycle = cyclical;
+  return cyclical;
+}
+
+/**
+ * Creates an object that references another object more
+ * than once in an acyclical way.
+ */
+function createDuplicateAcyclical () {
+  const duplicate = {};
```
Krinkle, line 28: To improve code coverage a little bit, and to weed out errors that result in objects rendering as empty without keys, perhaps include at least one key-value pair in here.

test/fixtures/unit.js
```diff
@@ -11,6 +11,27 @@ function copyErrors (testEnd) {
   return testEnd;
 }
 
+/**
+ * Creates an object that has a cyclical reference.
+ */
+function createCyclical () {
+  const cyclical = {};
+  cyclical.cycle = cyclical;
```
Krinkle, line 19: In order to excercise the stack a bit bit, perhaps add one layer of indirection here, e.g. `cyclical = { sub: {}; cyclical.sub.cycle = cyclical;`.

Krinkle (changes requested):

lib/reporters/TapReporter.js
zackthehuman: TIL what _idem_ means.

test/fixtures/unit.js
zackthehuman: Added an additional structure with a subobject cycle.

test/fixtures/unit.js
zackthehuman: Good point. Added keys to each of the objects so we can identify them in the output.

lib/reporters/TapReporter.js
zackthehuman: Fixed. Should I also avoid `Array#map`?

Krinkle (approved):

Issue #118, opened 2020-09-24
tape?


ljharb:
No idea what would be involved, but tape is more commonly used than most of the frameworks listed in the readme ¯\\\_(ツ)_/¯

Krinkle:
Sure thing, we wouldn't want a spec that is incompatible or inconvenient with tape.

I haven't been looking into additional frameworks for a while as I hope between the ones we have active integration with now, that we cover all the main primitives in this space. I'm currently focussing on publishing the spec and getting adoption from reporter plugins, and orchestrations/runners. But, the plugin ecosystem around Karma and QUnit hasn't been super active now, so (early) adoption is getting harder to bootstrap. Mocha certainly takes the cake on popularity, but tape is huge!

I went and created a team for tape here, invited a few folks, and made you team maintainer. Feel free to invite more people any time.

From a quick look around, I think tape's use of lexical binding for nested tests and assertions, would produce a similar model as QUnit. Is that right? Would love to learn more about any semantic differences, perhaps at https://github.com/js-reporters/js-reporters/issues/117.

There's also definitely room for more adapters to allow for early adoption, prior to any implementation shipping in tape itself, and/or a additional [research section](https://github.com/js-reporters/js-reporters/blob/main/docs/frameworks.md) on tape as a framework.

Welcome ❤️

ljharb:
I'm not familiar with QUnit at all, I'm afraid.

Krinkle:
Thanks for sticking with me, and sorry it took so long. It's odd that it seems in the years prior adopting TAP in this way was never considered. While support for it in Node.js tools has been widespread for a while, actually seeing it used in CLI output by default is still pretty rare apart form tape and node-tap, and using it for browser communication while proven to be viable indeed seems rare beyond testling. But, I love it, let's give it our best.

Considering this closing, continuing at https://github.com/js-reporters/js-reporters/issues/133.

PR #132, opened 2021-02-15
TapeAdapter: Add new adapter for tape

Krinkle:
Ref https://github.com/js-reporters/js-reporters/issues/118.

### DIFF
--- a/lib/adapters/TapeAdapter.js
+++ b/lib/adapters/TapeAdapter.js
@@ -23,7 +23,7 @@ module.exports = class TapeAdapter extends EventEmitter {
       if (!started) {
         started = true;
         this.emit('runStart', { name: null, counts: { total: null } });
-        runStartTime = new Date().getTime();
+        runStartTime = Date.now();
       }
     };
 
--- a/lib/adapters/TapeAdapter.js
+++ b/lib/adapters/TapeAdapter.js
@@ -36,7 +36,7 @@ module.exports = class TapeAdapter extends EventEmitter {
         // {type: test, name: '…', id: 1, skip: false, todo: false, parent: 0}
         case 'test':
           test = activeTests[row.id] = {
-            startTime: new Date().getTime(),
+            startTime: Date.now(),
             name: row.name,
             parentName: row.parent ? activeTests[row.parent].name : null,
             fullName: row.parent ? [...activeTests[row.parent].fullName, row.name] : [row.name],
--- a/lib/adapters/TapeAdapter.js
+++ b/lib/adapters/TapeAdapter.js
@@ -1,0 +1,134 @@
+const EventEmitter = require('events');
+
+module.exports = class TapeAdapter extends EventEmitter {
+  /**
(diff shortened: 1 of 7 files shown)


lib/adapters/TapeAdapter.js
```diff
@@ -0,0 +1,134 @@
...
+    // TODO: Unable to observe 'prerun' event for some reason.
+    let started = false;
+    let runStartTime;
+    let runStatus = 'passed';
+    const runCounts = { passed: 0, failed: 0, skipped: 0, todo: 0, total: 0 };
+    const startOnce = () => {
+      if (!started) {
+        started = true;
+        this.emit('runStart', { name: null, counts: { total: null } });
+        runStartTime = new Date().getTime();
```
ljharb, line 26:
```suggestion
        runStartTime = Date.now();
```

lib/adapters/TapeAdapter.js
```diff
@@ -0,0 +1,134 @@
...
+    let fakeTestId = -1;
+    const activeTests = {};
+    const handleRow = (row) => {
+      let test;
+      switch (row.type) {
+        // {type: test, name: '…', id: 0, skip: false, todo: false}
+        // {type: test, name: '…', id: 1, skip: false, todo: false, parent: 0}
+        case 'test':
+          test = activeTests[row.id] = {
+            startTime: new Date().getTime(),
```
ljharb, line 39:
```suggestion
            startTime: Date.now(),
```
same throughout; since 2011 `new Date().getTime()` has been a legacy pattern :-)

ljharb (commented):
It'd be good to add tests where the callback returns a Promise (both rejected and fulfilled), as well as tests that use `t.plan()` instead of `t.end()`.

lib/adapters/TapeAdapter.js
Krinkle: I'm aiming to support IE9+ in the long-term here since imho this should be a simple low-level tool that won't stand in the way, especially e.g. with browserstack, saucelabs, and karma integration.

Having said that, this part of ES5 indeed landed pretty early and widely, so I guess we can just use that. I intuitively avoided it knowing it got spec'ed by ES5, and hadn't really considered when browsers implemented it. Thanks!

ljharb: I’m a huge fan of supporting old browsers, but old browsers will almost certainly be polyfilled.

Krinkle: At what level though? At some point one can't "pass the buck" any further.

ljharb: i expect everything to have the equivalent of the es5-shim and es6-shim installed.

Krinkle:
Closing in favour of https://github.com/js-reporters/js-reporters/issues/133. I'll try to keep the js-reporters spec instead compatible with what existed before, and reduce its use to adapting framework toward TapReporter (instead of adapting from already-TAP-capable frameworks toward the CRI event model).

